PR #2, opened 2020-03-09
Creating database and user & granting user privileges to database

allopez7:


database.js
```diff
@@ -0,0 +1,28 @@
...
+const client = new Client({
+  host: process.env.HOST,
+  port: process.env.PORT,
+  user: process.env.USERNAME,
+  password: process.env.PASSWORD,
+  database: process.env.DATABASE 
+});
+
+client.connect().then(() => {
+  createPgAccount("username", "password", "database");
```
songz: what is this doing.

database.js
```diff
@@ -0,0 +1,28 @@
...
+    
+    const createUser = await client.query(`CREATE USER ${username} WITH ENCRYPTED password '${password}'`)
+
+    const userDatabasePrivileges = await client.query(`GRANT ALL PRIVILEGES ON DATABASE ${database} To ${username}`)
+
+    await client.end()
+  }catch(err){
+    console.log('error', err)
+  }
+};
```
songz: every create needs a delete. Please add a delete function.

database.js
allopez7: calling the function createPgAccount and passing in dummy data for now. Eventually, data from the sign-up page will get passed in.

songz: remove, or else this script will throw error.

database.js
songz: this file is missing tests. Could you write a test?

database.js
```diff
@@ -0,0 +1,28 @@
...
+
+const createPgAccount = async (username, password, database) => {
+  try{
+    const createDatabase = await client.query(`CREATE DATABASE ${database}`)
+    
+    const createUser = await client.query(`CREATE USER ${username} WITH ENCRYPTED password '${password}'`)
+
+    const userDatabasePrivileges = await client.query(`GRANT ALL PRIVILEGES ON DATABASE ${database} To ${username}`)
+
+    await client.end()
```
songz: why is your client ending when you've created 1 account? By doing this, you can't run this function more than once.

database.js
allopez7: Going to create another issue for that.

database.js
```diff
@@ -0,0 +1,28 @@
...
+  database: process.env.DATABASE 
+});
+
+client.connect().then(() => {
+  createPgAccount("username", "password", "database");
+});
+
+const createPgAccount = async (username, password, database) => {
+  try{
+    const createDatabase = await client.query(`CREATE DATABASE ${database}`)
```
hwong0305: Why do you need to assign the return the value to variables when it's not being used? 

It should just be `await client.query(CREATE DATABASE ${database}`)

allopez7: good point

database.js
```diff
@@ -0,0 +1,36 @@
...
+  host: process.env.HOST,
+  port: process.env.PORT,
+  user: process.env.USERNAME,
+  password: process.env.PASSWORD,
+  database: process.env.DATABASE 
+});
+
+client.connect().then(() => {
+  console.log('connected')
+});
```
songz: Put these inside a `startPgDB` function.

songz: Add comment 
```
// when the app starts, we should first start all the databases we support (ie PostGres, MongoDb, Neo4J). If the DB can't start up, then our service is useless, so don't even start the app.
```

package.json
```diff
@@ -0,0 +1,29 @@
...
+    "eslint-config-prettier": "^6.10.0",
+    "pg": "^7.18.2"
+  },
+  "devDependencies": {
+    "dotenv": "^8.2.0",
+    "jest": "^25.1.0"
+  },
+  "scripts": {
+    "test": "jest",
+    "start": "node database.js"
```
songz: `node index.js` <- usually the default

database/postgres/pg.js
```diff
@@ -0,0 +1,51 @@
...
+    client = new Client({
+      host: process.env.HOST,
+      port: process.env.PORT,
+      user: process.env.USERNAME,
+      password: process.env.PASSWORD,
+      database: process.env.DATABASE
+    })
+    await client.connect()
+  }catch(err){
+    console.log('connection failed', err)
```
songz, line 18: this should return an error so whomever calls `startPGDB` will know that it has failed.

songz:
```
return {
  error: {
    message: 'Connection to PG failed'
  }
}
```

songz: please add this as a test case.

database/postgres/pg.js
```diff
@@ -0,0 +1,51 @@
...
+  }
+}
+
+pgModule.closePGDB = async ()=>{
+  try{
+    await client.end()
+  }catch(err){
+    console.log('connection failed to closed', err)
```
songz, line 26: this should return an error so whomever calls `closePGDB` will know that it has failed.

songz: please add this as a test case

database/postgres/pg.js
```diff
@@ -0,0 +1,51 @@
...
+}
+
+pgModule.createPgAccount = async (username, password)=>{
+  try{
+    if(!username || !password) return
+    await client.query(`CREATE DATABASE IF NOT EXISTS ${username}`)
+    await client.query(`CREATE USER IF NOT EXISTS ${username} WITH ENCRYPTED password '${password}'`)
+    await client.query(`GRANT ALL PRIVILEGES ON DATABASE ${username} TO ${username}`)
+  }catch(err){
+    console.log('failed to createPgAccount', err)
```
songz, line 37: please throw an error here:

```
throw new Error(`failed to createPGAccount for user: ${username}`)
```

make sure your tests address this issue.

database/postgres/pg.js
```diff
@@ -0,0 +1,51 @@
...
+  }
+}
+
+pgModule.deletePgAccount = async (username)=>{
+  try{
+    if(!username) return
+    await client.query(`DROP DATABASE IF EXISTS ${username}`)
+    await client.query(`DROP USER IF EXISTS ${username}`)
+  }catch(err){
+    console.log('failed to deletePgAccount', err)
```
songz, line 47: please throw an error here. Make sure your test addresses this issue.

database/postgres/pg.js
```diff
@@ -0,0 +1,63 @@
...
+      error: {
+        message: 'Connection to PG failed to close'
+      }
+    }
+  }
+}
+
+pgModule.createPgAccount = async (username, password)=>{
+  try{
+    if(!username || !password) return
```
songz, line 42: move this out of try

database/postgres/pg.js
```diff
@@ -0,0 +1,63 @@
...
+    await client.query(`GRANT ALL PRIVILEGES ON DATABASE ${username} TO ${username}`)
+  }catch(err){
+    console.log('failed to createPgAccount', err)
+    throw new Error(`failed to createPgAccount for user: ${username}`)
+  }
+}
+
+pgModule.deletePgAccount = async (username)=>{
+  try{
+    if(!username) return
```
songz, line 54: move this out of try

database/postgres/pg.test.js
```diff
@@ -0,0 +1,61 @@
...
+  pgClient.Client = function(){
+    this.query = mockClient.query
+    this.connect = mockClient.connect
+    this.end = mockClient.end
+  }
+  let {startPGDB, closePGDB, createPgAccount, deletePgAccount} = require('./pg')
+  describe('Test startPGDB && closePGDB', ()=>{
+    it('startPGDB', async ()=>{
+      const startRes = await startPGDB()
+      if(startRes && startRes.error && startRes.error.message){
```
songz, line 21: remove `if` statement.

database/postgres/pg.js
```diff
@@ -0,0 +1,45 @@
...
+    await client.query(`CREATE USER IF NOT EXISTS ${username} WITH ENCRYPTED password '${password}'`)
+    await client.query(`GRANT ALL PRIVILEGES ON DATABASE ${username} TO ${username}`)
+  }catch(err){
+    console.log('failed to createPgAccount', err)
+    throw new Error(`failed to createPgAccount for user: ${username}`)
+  }
+}
+
+pgModule.deletePgAccount = async (username)=>{
+    if(!username) return
```
songz, line 35: missing test case. if you call `deletePgAccount()`, then `query` should not have been called.

database/postgres/pg.js
```diff
@@ -0,0 +1,45 @@
...
+    })
+    return client.connect()
+}
+
+pgModule.closePGDB = ()=>{
+    return client.end()
+}
+
+pgModule.createPgAccount = async (username, password)=>{
+    if(!username || !password) return
```
songz, line 23: missing test case. if you call `deletePgAccount()`, then `query` should not have been called.

database/postgres/pg.test.js
```diff
@@ -0,0 +1,55 @@
...
+      expect(mockClient.connect).toHaveBeenCalledTimes(1)
+    })
+    it('closePGDB', async ()=>{
+      await closePGDB()
+      expect(mockClient.end).toHaveBeenCalledTimes(1)
+    })
+  })
+  describe('Test createPgAccount', ()=>{
+    it('createPgAccount', async ()=>{
+      await startPGDB()
```
songz, line 30: everything is mocked, you don't need this.

songz: nvm u need it to initialized client.

database/postgres/pg.test.js
```diff
@@ -0,0 +1,55 @@
...
+    it('createPgAccount', async ()=>{
+      await startPGDB() 
+      await createPgAccount('username', 'password')
+      expect(mockClient.query).toHaveBeenCalledTimes(3)
+      expect(mockClient.query).toHaveBeenNthCalledWith(1, `CREATE DATABASE IF NOT EXISTS username`)
+      expect(mockClient.query).toHaveBeenNthCalledWith(2, `CREATE USER IF NOT EXISTS username WITH ENCRYPTED password 'password'`)
+      expect(mockClient.query).toHaveBeenNthCalledWith(3, `GRANT ALL PRIVILEGES ON DATABASE username TO username`)
+      expect(async ()=>{
+        await createPgAccount()
+      }).not.toThrow()
```
songz, line 38: expecting it to not throw is not doing anything. because your test doesn't have try catch. If the function threw something, it would not pass

database/postgres/pg.test.js
```diff
@@ -0,0 +1,55 @@
...
+  describe('Test deletePgAccount', ()=>{
+    it('deletePgAccount', async ()=>{
+      await startPGDB()
+      await deletePgAccount('username')
+      expect(mockClient.query).toHaveBeenCalledTimes(2)
+      expect(mockClient.query).toHaveBeenNthCalledWith(1, `DROP DATABASE IF EXISTS username`)
+      expect(mockClient.query).toHaveBeenNthCalledWith(2, `DROP USER IF EXISTS username`)
+      expect(async ()=>{
+        await deletePgAccount('username')
+      }).not.toThrow()
```
songz, line 51: expecting it to not throw is not doing anything. because your test doesn't have try catch. If the function threw something, it would not pass

database/postgres/pg.test.js
```diff
@@ -0,0 +1,61 @@
...
+      await startPGDB()
+      expect(mockClient.connect).toHaveBeenCalledTimes(1)
+    })
+    it('closePGDB', async ()=>{
+      await closePGDB()
+      expect(mockClient.end).toHaveBeenCalledTimes(1)
+    })
+  })
+  describe('Test createPgAccount', ()=>{
+    it('createPgAccount', async ()=>{
```
songz, line 29: should run all necessary queries in a successful createPgAccount call

database/postgres/pg.test.js
```diff
@@ -0,0 +1,61 @@
...
+        await startPGDB() 
+        await createPgAccount('username', 'password')
+        expect(mockClient.query).toHaveBeenCalledTimes(3)
+        expect(mockClient.query).toHaveBeenNthCalledWith(1, `CREATE DATABASE IF NOT EXISTS username`)
+        expect(mockClient.query).toHaveBeenNthCalledWith(2, `CREATE USER IF NOT EXISTS username WITH ENCRYPTED password 'password'`)
+        expect(mockClient.query).toHaveBeenNthCalledWith(3, `GRANT ALL PRIVILEGES ON DATABASE username TO username`)
+        await closePGDB()
+    })
+    it('createPgAccount, undefined parameters', async ()=>{
```
songz, line 38: it should not execute any queries if required arguments are not passed in.

database/postgres/pg.test.js
```diff
@@ -0,0 +1,61 @@
...
+  describe('Test deletePgAccount', ()=>{
+    it('deletePgAccount', async ()=>{
+        await startPGDB()
+        await deletePgAccount('username')
+        expect(mockClient.query).toHaveBeenCalledTimes(2)
+        expect(mockClient.query).toHaveBeenNthCalledWith(1, `DROP DATABASE IF EXISTS username`)
+        expect(mockClient.query).toHaveBeenNthCalledWith(2, `DROP USER IF EXISTS username`)
+        await closePGDB()
+    })
+    it('deletePgAccount, undefined parameters', async ()=>{
```
songz, line 54: it should not execute any queries in deletePgAccount if required arguments are not passed in.

database/postgres/pg.test.js
```diff
@@ -0,0 +1,61 @@
...
+    })
+    it('createPgAccount, undefined parameters', async ()=>{
+        await startPGDB()
+        await createPgAccount()
+        expect(mockClient.query).toHaveBeenCalledTimes(0)
+        await closePGDB()
+    })
+  })
+  describe('Test deletePgAccount', ()=>{
+    it('deletePgAccount', async ()=>{
```
songz, line 46: it should execute all queries if required arguments are passed into deletePgAccount

database/postgres/pg.test.js
```diff
@@ -0,0 +1,61 @@
...
+    end: jest.fn().mockReturnValue(Promise.resolve())
+  }
+  pgClient.Client = function(){
+    this.query = mockClient.query
+    this.connect = mockClient.connect
+    this.end = mockClient.end
+  }
+  let {startPGDB, closePGDB, createPgAccount, deletePgAccount} = require('./pg')
+  describe('Test startPGDB && closePGDB', ()=>{
+    it('startPGDB', async ()=>{
```
songz, line 19: should call connect when starting PGDB

database/postgres/pg.test.js
```diff
@@ -0,0 +1,61 @@
...
+      await startPGDB()
+      expect(mockClient.connect).toHaveBeenCalledTimes(1)
+    })
+    it('closePGDB', async ()=>{
```
songz, line 23: should call end when starting PGDB

database/postgres/pg.test.js
```diff
@@ -0,0 +1,61 @@
+const pgClient = require('pg')
```
songz, line 1:
```
jest.mock('pg')
const {Client} = require('pg')
const {startPGDB.....} = require('./pg')
```

database/postgres/pg.test.js
```diff
@@ -0,0 +1,61 @@
...
+  let mockClient = {
+    query: jest.fn().mockReturnValue(Promise.resolve()),
+    connect: jest.fn().mockReturnValue(Promise.resolve()),
+    end: jest.fn().mockReturnValue(Promise.resolve())
+  }
+  pgClient.Client = function(){
+    this.query = mockClient.query
+    this.connect = mockClient.connect
+    this.end = mockClient.end
+  }
```
songz, line 16: in addition to comment (above), you can do this to be more concise:
```
Client.mockImplementation(function() {
  return mockClient
})
```

database/postgres/pg.test.js
```diff
@@ -0,0 +1,80 @@
+jest.mock('pg')
+const {Client} = require('pg')
+const {startPGDB, closePGDB, createPgAccount, deletePgAccount} = require('./pg')
+
+describe('Test PG DB', ()=>{
+  afterEach(()=>{
```
kimong1, line 6: This should be `beforeEach(...)` shoudn't it?

database/postgres/pg.test.js
```diff
@@ -0,0 +1,80 @@
...
+  })
+  let mockClient = {
+    query: jest.fn().mockReturnValue(Promise.resolve()),
+    connect: jest.fn().mockReturnValue(Promise.resolve()),
+    end: jest.fn().mockReturnValue(Promise.resolve())
+  }
+  Client.mockImplementation(function(){
+    return mockClient
+  }) 
+  global.console.log = jest.fn()
```
kimong1, line 17: I think it's `console.log = jest.fn()`, also it'll be clearer if it moves to under 47th line

database/postgres/pg.test.js
allopez7: It could be `beforeEach` as well but on the first test the mocks would be cleared when there is no need to clear.

database/postgres/pg.test.js
allopez7: but that wouldn't make sense for line 75 where it's called again. Since it's global i'm changing the functionality of `console` so I wanted to make it clear that this is changing the functionality of console for the whole database file. You do you have a good point so I'm going to take it outside of the postgres test so we can use it in the MongoDB and Neo4J test as well.

database/postgres/pg.test.js
songz: should be `beforeEach`. As the test suite gets more complex, this will cause issues later on, you can't rely on other ppl to always clear their test.

database/postgres/pg.test.js
songz: this should be scoped into the function itself. Its not good for a developer later on to try to `console.log` something, but nothing shows up. This will essentially remove `console.log` functionality, which will remove debugging feature for alot of devs. Better to be inside the test.

database/postgres/pg.test.js
```diff
@@ -0,0 +1,80 @@
...
+    })
+  })
+  describe('Test create and delete pgAccount', ()=>{
+    beforeEach(async ()=>{
+      await startPGDB()
+    })
+    afterEach(async ()=>{
+      await closePGDB()
+    })
+    describe('Test createPgAccount', ()=>{
```
songz, line 35: delete this `describe` grouping, there are too many nested describes.

PR #19, opened 2020-04-03
mailgun takes email input to send confirmation email template

pvalderama:


app.js
```diff
@@ -0,0 +1,36 @@
+const express = require('express');
+const Mailgun = require('mailgun-js')
+
+const app = express();
+
+const api_key = '317a0f64411bd75d7598da6be916803b-46ac6b00-50f5a2ee';
```
allopez7, line 6: You will need to import from .env file, which is written like this require('dotenv').config(). Inside of .env make sure you create API_KEY= '317....'

app.js
```diff
@@ -0,0 +1,36 @@
...
+
+const domain = 'sandbox47881c22500a4bf7a21f621e5bc25800.mailgun.org';
```
allopez7, line 8: also add domain to .env file as DOMAIN='sandbox....'

allopez7, line 8: this can also be added to .env to make things look cleaner

app.js
```diff
@@ -0,0 +1,36 @@
...
+
+const from_who = 'pvalderama@gmail.com';
```
allopez7, line 10: and this

app.js
```diff
@@ -0,0 +1,36 @@
...
+    const fp = path.join(__dirname, 'template.txt')
+
+    const mailgun = new Mailgun({apiKey: api_key, domain: domain});
+
+    const data = {
+        from: from_who,
+        to: req.params.mail,
+        subject: 'Confirm E-mail for C0D3',
+        text: 'Confirm E-mail',
+        attachment: fp
```
allopez7, line 24: fp could be more descriptive

app.js
songz: `process.env.MAIL_KEY`

songz: `./lib/mailer.js`

PR #24, opened 2020-04-08
Logging module

kimong1:


log.js
```diff
@@ -0,0 +1,23 @@
...
+
+const logger = winston.createLogger({
+    transports : [
+        new winston.transports.Console()
+    ]
+})    
+
+const output = (file) => {
+    const obj = {}
+    obj.log = (str) => {
```
songz, line 11:
```
(...args) => logger.info(file, ...args)
```

Same with all the functions. and then you need `log.test.js`

kimong1: I jsut tested the first part. When I do `logger.info(file, args)`, it prints out `{"level": "info", "message": file}` and ignores `args`

> ```
> (...args) => logger.info(file, ...args)
> ```
> 
> Same with all the functions. and then you need `log.test.js`

package.json
```diff
@@ -0,0 +1,25 @@
+{
```
allopez7, line 1: You need to git pull the current changes otherwise you will overwrite package.json

package-lock.json
```diff
@@ -0,0 +1,269 @@
+{
```
allopez7, line 1: Do you push this file up.

songz: this should always be pushed up.

lib/log.test.js
```diff
@@ -0,0 +1,23 @@
+jest.mock('winston')
+const {createLogger} = require('winston')
+jest.unmock('winston')
```
songz, line 3: why?

lib/log.test.js
```diff
@@ -0,0 +1,23 @@
+jest.mock('winston')
+const {createLogger} = require('winston')
```
songz, line 2:
```
const winston = require('winston')
```

lib/log.test.js
```diff
@@ -0,0 +1,23 @@
...
+jest.unmock('winston')
+const {format, transports} = require('winston')
```
songz, line 4: remove this line. its not used anywhere.

lib/log.test.js
```diff
@@ -0,0 +1,23 @@
...
+const logger = require('./log')(__filename)
+
+describe('Testing log.js', () => {
+    beforeEach(() => {
+        jest.clearAllMocks()
```
songz, line 9: `winston.createLogger = jest.fn()`

lib/log.test.js
```diff
@@ -0,0 +1,23 @@
...
+    })
+    it('should call createLogger when logger.error is called', () => {
+        logger.error('test message for error')
+        expect(createLogger).toHaveBeenCalled()
```
songz, line 13:
```
expect(winston.createLogger).toHaveBeenCalled()
```

lib/log.test.js
kimong1: Without this line it throw error because `./log` uses those function which are from `'winston'` but it's already mocked.

lib/log.test.js
```diff
@@ -0,0 +1,27 @@
...
+    }
+})
+winston.format = {
+    combine: () => {},
+    timestamp: () => {},
+    label: () => {},
+    colorize: () => {},
+    printf: () => {}
+}
+winston.transports.Console = function(){}
```
kimong1, line 17: It works but doesn't look efficient. I want to use `jest.requireActual()` [tutorial](https://jestjs.io/docs/en/bypassing-module-mocks)
If anyone has idea why it didn't work it'll be helpful

lib/log.test.js
```diff
@@ -0,0 +1,27 @@
...
+
+describe('Testing log.js', () => {
+    beforeEach(() => {
+        jest.clearAllMocks()
+    })
+    it('should call winston.createLogger when log is required', () => {
+        const logger = require('./log')(__filename)
+        expect(winston.createLogger).toHaveBeenCalled()
+    })
```
kimong1, line 26: `createLogger` is called only once when `logger` is defined in `./log` so I changed this part to this from when there were 3 tests that are for `log.error`, `log.warn`, `log.info`

lib/log.test.js
songz: you need to remove `jest.mock('winston')` to use the actual `winston` library.

Then you do....
```
winston.createLogger = jest.fn()
const logger = require('./log')(__filename)
expect(winston.createLogger).toHaveBeenCalled()
```

PR #28, opened 2020-04-18
Sequelize setup 

allopez7:


sequelize/dbQueries.js
```diff
@@ -0,0 +1,45 @@
+const {Account} = require('./db')
```
allopez7, line 1: did not intend to push this up.

Issue #27, opened 2020-04-16
[Setup] - Create env example file


songz:
Create `.env.example` and with sample information and comments about what each env variable is used for.

1. Update readme to move `.env.example` to `.env` and what accounts users have to create / setup to get the variables needed.
2. Update `.gitignore` to ignore `.env` files so nobody will accidentally commit that file.

songz:
Done

PR #37, opened 2020-04-23
#27: creates env.example and fixed pg.js

kimong1:
`env.example` explains env variables that are used in this project.

I added `env.example` file with env variables that are used in `pg.js`
Also fixed `pg.js` since I realized a typo in the variable name(10th line)

### DIFF
--- a/env.example
+++ b/env.example
@@ -1,0 +1,5 @@
+HOST=
+PORT=3000
+USERNAME=username
+PASSWORD=password
+DATABASE=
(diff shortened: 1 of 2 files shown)


env.example
```diff
@@ -0,0 +1,5 @@
+HOST=
```
allopez7, line 1: you can add a dummy domain here

env.example
```diff
@@ -0,0 +1,5 @@
...
+PORT=3000
+USERNAME=username
+PASSWORD=password
+DATABASE=
```
allopez7, line 5: you can call it database

Issue #31, opened 2020-04-20
[logger] - Allow object input or error input


songz:
`logger.error` should accept error object, objects, and/or string.

All the other logger functions should allow for objects, strings, array, or other data types.

If it is `error`, you can call the `.toString()` method.

https://stackoverflow.com/questions/30469261/checking-for-typeof-error-in-js

`JSON.stringify( [1,2,3], null, 2)`

songz:
Completed

PR #38, opened 2020-04-24
#31: updates logging module to allow multiple arguments and data type…

kimong1:
#31: updates logging module to allow multiple arguments and data types that are not string
------------------------------------------------------------------------------------------------------
This update allows followings:
  * Logging module now takes in multiple arguments and it logs each of them in each line.
  * Also it takes in data types like object, array, etc. as well as string.


lib/log.js
```diff
@@ -16,5 +16,24 @@ module.exports = (file) => {
             new transports.Console()
         ]
     })
-    return logger        
+
+    const obj = {
+        info: (...args) => {
+            args.forEach((e) => {
+                logger.info(JSON.stringify(e))
```
songz, line 23: If I pass in 5 arguments, I wouldn't want to log 5 times. I think its better to log 1 time as a long string.

1. Helper function that maps through `args`. If the element type is Error, then return `e.toString()`. Otherwise, return the element
2. The helper function returns the string generated from `util.format('%j', newArray)`. I use `util.format` because I think `JSON.stringify puts too many quotes, makes it hard to see.`
3. Then I will just log once: `logger.info( processArgs(args) )`

lib/log.js
```diff
@@ -16,5 +16,24 @@ module.exports = (file) => {
...
+                logger.warn(JSON.stringify(e))
+            })
+        },
+        error: (...args) => {
+            args.forEach((e) => {
+                logger.error(JSON.stringify(e))
+            })
+        }
+    }
+
```
songz, line 37: I think you should add some test to your logger.

lib/log.js
```diff
@@ -16,5 +17,30 @@ module.exports = (file) => {
             new transports.Console()
         ]
     })
-    return logger        
+
+    const obj = {
+        info: (...args) => {
+            const mappedArgs = args.map((e) => {
```
songz, line 23: make this a helper function:
```
const processArgs = (args) => {
  return args.map....
}
```

Issue #16, opened 2020-03-27
Email Module


songz:
# Goal
As a developer, I want to be able to run a `sendEmail` function with an email and a type. The email function should take care of sending the email.

Email should also generate email Template Strings.

# Requirements
```
const {generateTemplate, sendEmail } = require('./email')
sendEmail( 'hello@example.com', 'subject string', generateTemplate('confirmation', data) )
```

# TODO
1. Get Mailgun account, use their API to send an email successfully.
2. sendEmail Function
3. GenerateTemplate Function

songz:
merged

PR #41, opened 2020-05-06
#16: fixes mailer and test

kimong1:
Fixes mailer and test

### DIFF
--- a/services/mailer.test.js
+++ b/services/mailer.test.js
@@ -1,0 +1,45 @@
+jest.mock('../lib/log')
+const logGen = require('../lib/log')
+const logger = {
+    info: jest.fn(),
+    error: jest.fn()
+}
+logGen.mockReturnValue(logger)
+
+jest.mock('mailgun-js')
+const mailgun = require('mailgun-js');
+const messages = {
+    send: jest.fn().mockImplementation((data) => {
+        if (data.to) return Promise.resolve('hello')
+        return Promise.reject('rejected')
+    })
+}
+mailgun.mockImplementation(() => {
+    return {
+        messages: () => {
+            return messages
+        }
+    }
+})
+const email = require('./mailer');
+
+describe('Test mailgun', ()=>{
+    beforeEach(() => {
+        jest.clearAllMocks()
+    })
+
+    it('should test if mocksend and mailgun is called', async () => {
+        await email.sendConfirmationEmail('paul@github.com', 'token123')
+        expect(messages.send).toHaveBeenCalledTimes(1)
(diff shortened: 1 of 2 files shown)


services/mailer.test.js
```diff
@@ -0,0 +1,45 @@
...
+const logger = {
+    info: jest.fn(),
+    error: jest.fn()
+}
+logGen.mockReturnValue(logger)
+
+jest.mock('mailgun-js')
+const mailgun = require('mailgun-js');
+const messages = {
+    send: jest.fn().mockImplementation((data) => {
```
songz, line 12: I think a better way is to us pointers.
```
const messages = {
  send: jest.fn().mockReturnValue(Promise.resolve('hello'))
}
mailgun.mockImplementation( () => {
  return {messages}
})
```

services/mailer.test.js
```diff
@@ -0,0 +1,45 @@
...
+    it('should test if mocksend and mailgun is called', async () => {
+        await email.sendConfirmationEmail('paul@github.com', 'token123')
+        expect(messages.send).toHaveBeenCalledTimes(1)
+        expect(messages.send.mock.calls[0][0]).toMatchSnapshot()
+        expect(logger.info).toHaveBeenCalledTimes(1)
+        expect(logger.info.mock.calls[0][0]).toEqual('Confirmation Email successfully sent')
+    })
+
+    it('should call logger.error when function is called with invalid argument', async () => {
+        await email.sendConfirmationEmail(null, null)
```
songz, line 40: you should mock reject, I think its important to test here when service is down. like:
```
messages.send = jest.fn().mockReturnValue( Promise.reject('invalid') )
```

services/mailer.test.js
```diff
@@ -0,0 +1,45 @@
...
+})
+const email = require('./mailer');
+
+describe('Test mailgun', ()=>{
+    beforeEach(() => {
+        jest.clearAllMocks()
+    })
+
+    it('should test if mocksend and mailgun is called', async () => {
+        await email.sendConfirmationEmail('paul@github.com', 'token123')
```
songz, line 32: read comment below.
```
messages.send = jest.fn().mockReturnValue( Promise.resolve('hello') )
```

services/mailer.test.js
kimong1: What is pointers?

songz: non-primitive data types?

services/mailer.test.js
kimong1: I think it does the same as what it does in 13-14 lines. I see how it's easier to be read. Is that what you meant?

Issue #32, opened 2020-04-20
[Backend Functions] - Signup


songz:
helper functions.

`const signup = (userObj) => {`

`signup` should take in `email`, `username`, `password`. 
* Check validation rules and return an object with errors. 
  * `username` contains invalid characters (must be alpha numeric)
  * `email` is invalid
  * `username` / `password` is too short / long
* Call sequelize's database function for signup
  * If error happens, throw error.
* Call mailgun function to send email
* Respond with user object.

PR #43, opened 2020-05-09
#32 Signup validation rules

greenc123:


lib/users.js
```diff
@@ -0,0 +1,19 @@
+let yup = require('yup')
```
songz, line 1: `const`

lib/users.js
```diff
@@ -0,0 +1,19 @@
...
+
+const users = {}
+
+const schema = yup.object().shape({
+  username: yup.string().required().min(1).matches(/^[0-9a-zA-Z]+$/, "username cannot contain non-alphanumeric characters"),
```
songz, line 6: `.min(1)` <- remove

sequelize/db.js
```diff
@@ -60,4 +62,9 @@ dbModule.close = ()=>{
   return sequelize.close()
 }
 
+dbModule.getModels = ()=>{
+  console.log('getting models', models)
```
songz, line 66: remove.

PR #44, opened 2020-05-12
Signup validation

greenc123:


sequelize/db.js
```diff
@@ -5,19 +5,21 @@ require('dotenv').config()
 let sequelize
 const dbModule = {}
 
-dbModule.start = async ()=>{
+const models = {}
+
+dbModule.start = async () => {
   logger.log('starting sequelize server')
```
kimong1, line 11: logger doesn't have log property. How about using logger.info?

sequelize/db.test.js
```diff
@@ -9,22 +9,22 @@ const logger = {
...
-    authenticate: jest.fn().mockReturnValue(Promise.resolve()),
-    define: jest.fn(),
-    sync: jest.fn().mockReturnValue(Promise.resolve()),
-    close: jest.fn().mockReturnValue(Promise.resolve())
-  }
-  Sequelize.mockImplementation(function(){
-    return mockSequelize
-  })
   it('should test how many times authenticate is called', ()=>{
     expect(logger.log).toHaveBeenCalledTimes(3)
```
kimong1, line 29: Lots of `logger.log` is used and it seems like it's been here before. How did it pass the test?

kimong1, line 29: `logger.log` passed test because it's mocked at the beginning

PR #46, opened 2020-05-17
Adds login function

kimong1:
I just added `logIn` function property to `users`
It includes valdation / encryption then returns user data.
Tests are ready.
Please review.


lib/users.js
```diff
@@ -8,12 +11,45 @@ const schema = yup.object().shape({
...
+users.signUp = async (userInfo) => {
   return schema.validate(userInfo).catch((err) => { return {
       errors: err.errors
     }
   })
 } 
 
+users.logIn = async (userInfo) => {
+  try {
+    await schema.validate(userInfo)
```
songz, line 23: remove. User input may be a username or email.

songz: you can just do a simple check:

```
if (!userInfo.username && !userInfo.email) {
  logger.info('invalid userInfo for login', userInfo)
  return {
    errors: ['invalid input']
  }
```

lib/users.js
```diff
@@ -8,12 +11,45 @@ const schema = yup.object().shape({
...
+  }
+  catch (err) {
+    logger.info(err)
+    return {
+      errors: err.errors
+    }
+  }
+
+  db.start()
```
songz, line 32: remove `db.start()`

lib/users.test.js
```diff
@@ -50,3 +50,79 @@ describe('Validates user signup', () => {
     })
   })
 })
+
+jest.mock('../sequelize/db')
+const bcrypt = require('bcrypt')
+const db = require('../sequelize/db')
```
songz, line 56: move these to the top of the file. Other user functions may need to use these. All mocks and requires must be at the top of the file.

lib/users.test.js
```diff
@@ -50,3 +50,79 @@ describe('Validates user signup', () => {
...
+
+const genData = async (password) => {
+  return {
+    username: 'testuser',
+    password: await bcrypt.hash(password, 10),
+    email: 'test.user@databases.com'
+  }
+}
+
+db.start = () => {}
```
songz, line 66: remove

lib/users.test.js
```diff
@@ -50,3 +50,79 @@ describe('Validates user signup', () => {
...
+const mockFindOne = jest.fn()
+db.getModels = () => {
```
songz, line 68: move the db mocks to the top of the file.

lib/users.test.js
```diff
@@ -50,3 +50,79 @@ describe('Validates user signup', () => {
...
+    }
+}
+
+const logIn = users.logIn
+
+beforeEach(() => {
+    jest.clearAllMocks()
+})
+
+describe('Validates user logIn', () => {
```
songz, line 82: This doesn't deserve to be its on suite (every `describe` is a test suite)

lib/users.test.js
```diff
@@ -50,3 +50,79 @@ describe('Validates user signup', () => {
...
+  it("function should reject if username doesn't exist", () => {
```
songz, line 83: `should reject if user input is invalid`

lib/users.test.js
```diff
@@ -50,3 +50,79 @@ describe('Validates user signup', () => {
     })
   })
 })
+
+jest.mock('../sequelize/db')
+const bcrypt = require('bcrypt')
+const db = require('../sequelize/db')
+
+const genData = async (password) => {
```
greenc123, line 58: Instead of genData could you rename it to genTestUserDataWithPassword

greenc123: genData is kind of confusing

lib/users.js
kimong1: That's a good idea

lib/users.test.js
```diff
@@ -1,6 +1,31 @@
...
+        Accounts: {
+            findOne: mockFindOne
+        }
+    }
+}
 
 const signUp = users.signUp
+const logIn = users.logIn
+
+beforeEach(() => {
```
songz, line 26: this belongs inside each test suite (aka `describe`)

PR #52, opened 2020-05-22
sign up functionality and user confirmation email

greenc123:


greenc123:
Closed for PR #54 instead

Issue #40, opened 2020-04-28
[Backend Function] - Send user password reset link on signup


songz:
`/lib/user.js` should have a function called `sendPwReset`

* Takes in a user object model
* Creates a new token. Maybe [crypto.randomBytes](https://nodejs.org/api/crypto.html#crypto_crypto_randombytes_size_callback)
* JSON stringify this: `{userId: id, userToken: token}`, then base64encode it.
* Sends Confirmation Email with the encoded data above
* Updates account token and expiration date

songz:
Outdated. Replaced by signup functionality

PR #54, opened 2020-05-23
Issues #40 and #32 email confirmation, password reset, signup funcationality and mailgun

greenc123:


services/mailer.js
```diff
@@ -1,32 +1,32 @@
...
 
-mgModule.sendConfirmationEmail = (receiver, token) => {    
-    const link = `https://learndatabases.dev/emailConfirmation/${token}`
-    const data = {
-        from: 'admin@learndatabases.dev',
-        to: receiver,
-        subject: 'Congratulations!',
-        text: 'Welcome to C0D3',
-        html: `
+mgModule.sendPasswordResetEmail = (receiver, token) => {
```
kimong1, line 9: Since this name has changed, one in the test should be changed too.

songz: good catch

songz: @greenc123 https://github.com/garageScript/databases/blob/master/services/mailer.test.js#L29

Issue #35, opened 2020-04-20
[Backend Functions] - Set user DB PW


songz:
Update user account `lib/user.js`
* setDBPassword that takes in an argument, and updates the Account sequelize table.

PR #57, opened 2020-05-26
#35 adds setDBPassword function

kimong1:
It adds `setDBPassword` function to `users` object.
Test is ready too.
Please review.

### DIFF
--- a/lib/users.js
+++ b/lib/users.js
@@ -63,6 +63,10 @@ users.logIn = async (userInfo) => {
 }
 
 users.setDBPassword = async (userInfo) => {
+  if (!userInfo.dbPassword || !userInfo.username) {
+    logger.info('invalid input: username and DB password cannot be empty', userInfo.username)
+    throw new Error('invalid input: username and DB password cannot be empty')
+  }
   const {Accounts} = db.getModels()
   try {
     Accounts.update({
--- a/lib/users.js
+++ b/lib/users.js
@@ -62,4 +62,21 @@ users.logIn = async (userInfo) => {
   return account.dataValues
 }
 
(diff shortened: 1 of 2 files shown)


lib/users.test.js
```diff
@@ -13,17 +20,20 @@ const genUserInfoWithPw = async (password) => {
...
-            create: mockCreateAccount
+            create: mockCreateAccount,
+            update: mockUpdate
         }
     }
 }
 
 const signUp = users.signUp
 const logIn = users.logIn
+const setDBPassword = users.setDBPassword
```
greenc123, line 36: This could all be done on the same line using destructuring. 
`const {signUp, logIn, setDBPassword} = users

lib/users.js
```diff
@@ -62,4 +62,21 @@ users.logIn = async (userInfo) => {
   return account.dataValues
 }
 
+users.setDBPassword = async (userInfo) => {
```
songz, line 65: could you add a check here? 
```
const {dbPassword, username} = userInfo
if (!dbPassword | | !username) {
  throw new Error('Invalid input: username and password cannot be empty')
}
```

lib/users.test.js
```diff
@@ -1,3 +1,12 @@
+jest.mock('./log')
+const logGen = require('./log')
+const logger = {
+    info: jest.fn(),
+    error: jest.fn()
+}
+logGen.mockReturnValue(logger)
+const {signUp, logIn, setDBPassword} = require('./users')
```
greenc123, line 8: Shouldn't `users = require('./users')` and `const {signUp, logIn, setDBPassword} = users` be on separate lines?

Issue #50, opened 2020-05-19
[API] - Signup / delete user


songz:
Make sure to follow rest conventions

```
Signup:
POST /api/users

Delete Users:
DELET /api/users/:id
```

PR #66, opened 2020-05-30
#50 Create / Delete User API

kimong1:
Create / Delete User API


kimong1:
`server.js` did not pass 100% coverage
image
This uncovered code will be tested in another PR(#68 )
Conflicts will be resolved after the other PR is merged

src/userRoutes.js
```diff
@@ -0,0 +1,58 @@
...
+    try {
+        const account = await Accounts.findOne({
+        where: {
+            id: req.params.id
+        }
+        })
+        if (!account) {
+        logger.info("Cannot find user", req.params.id)
+        res.status(500).json({error: {message: "Cannot find user"}})
+        return
```
greenc123, line 42: I think we should be returning the res so our functions don't return undefined for unintended consequences

`return res.status(500).json({error: {message: "Cannot find user"}})`

src/userRoutes.js
```diff
@@ -0,0 +1,58 @@
...
+        }
+        if (account.username !== req.session.username) {
+            logger.error("Username does not match to cookie", account.username, req.session.username)
+            res.status(500).json({error: {message: "Username does not match to cookie"}})
+            return
```
greenc123, line 47: Same thing here

src/userRoutes.js
songz: it doesn't matter

src/userRoutes.test.js
```diff
@@ -0,0 +1,121 @@
...
+    })
+    it('should send error if input is invalid', async () => {
+        const req = {
+            body: {
+                username: null,
+                email: null,
+                password: null
+            }
+        }
+        await createUser(req, res)
```
greenc123, line 34: Test for res.status

src/userRoutes.test.js
```diff
@@ -0,0 +1,121 @@
...
+            return
+        })
+        const req = {
+            body: {
+                username: 'username',
+                email: 'em@i.l',
+                password: '1q2w3e4r'
+            }
+        }
+        await createUser(req, res)
```
greenc123, line 62: test for res.status to be 200 if success

src/userRoutes.test.js
```diff
@@ -0,0 +1,121 @@
...
+    it('should send error if user session does not match', async () => {
+        const req = {
+            params: { id: 99999999 },
+            session: { username: 'testuserA' }
+        }
+        mcokFindOne.mockReturnValue({
+            username: 'testuserB'
+        })
+        await deleteUser(req, res)
+        return expect(res.json.mock.calls[0][0].error.message).toEqual("Username does not match to cookie")
```
greenc123, line 95: Test res.status for 500 if username does not match cookie

src/routes/userRoutes.js
```diff
@@ -35,5 +35,56 @@ routes.resetPassword = async (req, res) => {
...
 
+routes.createUser = async (req, res) => {
+  const userInfo = {
+      username: req.body.username,
+      email: req.body.email,
+      password: req.body.password
+  }
+  if (!userInfo.username || !userInfo.email || !userInfo.password) {
+      logger.info("invalid input")
+      return res.status(400).json({error: {message: "invalid input"}})
```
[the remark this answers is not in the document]
kimong1: Nice catch

src/routes/userRoutes.test.js
```diff
@@ -3,36 +3,41 @@ jest.mock('../../lib/users')
...
+  }
+}
 
 const res = {
-  status: jest.fn().mockReturnValue({json: mockJson})
+  status: jest.fn().mockImplementation(() => {
+    return res
+  }),
+  json: jest.fn()
 }
```
AliceHuang1027, line 23: i actually changed my code  
`status: jest.fn().mockImplementation(() => {
     return res
   }),
   json: jest.fn()` to 
`status: jest.fn().mockReturnValue({json: mockJson}) ` 😞

AliceHuang1027: we shall make an agreement to which way to choose

src/routes/userRoutes.test.js
```diff
@@ -50,11 +55,7 @@ describe('Testing user routes', () => {
       email: 'hello@world.com'
     }
 
-    db.getModels = jest.fn().mockReturnValue({
-      Accounts: {
-        findOne: jest.fn().mockReturnValue(userAccount)
-      }
-    })
+    mcokFindOne.mockReturnValue(userAccount)
```
AliceHuang1027, line 58: same as here:
i changed `mcokFindOne.mockReturnValue(userAccount)`
to
`  db.getModels = jest.fn().mockReturnValue({
       Accounts: {
         findOne: jest.fn().mockReturnValue(userAccount)
       }
     })
` let's make an agreement. i think both way worked

src/routes/userRoutes.test.js
```diff
@@ -68,7 +69,7 @@ describe('Testing user routes', () => {
 
     await resetPassword(req, res)
     expect(res.status.mock.calls[0][0]).toEqual(500)
-    expect(mockJson.mock.calls[0][0]).toEqual({ error: {message: 'Email delivery failed. Please try again'}})
+    expect(res.json.mock.calls[0][0]).toEqual({ error: {message: 'Email delivery failed. Please try again'}})
```
AliceHuang1027, line 72: same here

src/server.test.js
```diff
@@ -1,81 +1,74 @@
...
+const app = {
+  use: () => {},
+  get: () => {},
+  post: jest.fn(),
+  delete: jest.fn(),
+  listen: jest.fn().mockImplementation((port, callback) => callback()),
+  name: 'Carl Sagan'
+}
+express.mockReturnValue(app)
+
```
AliceHuang1027, line 26: damn it. i also comprised the way carlos wrote it and changed my code to the reverse way

src/routes/userRoutes.test.js
kimong1: We want `status` to be mocked as well. Because we want to see what argument is being passed to `status` function. Does it make sense?

src/routes/userRoutes.test.js
kimong1: Both way works. I prefer the way you did. But only in this line, it was more convinient to me to set a container called `mockFindOne` because it is re-mocked later

src/routes/userRoutes.test.js
kimong1: I like `res.json` better because it looks the same how the tested code does. I think this way is easier for the reviewers tounderstand the code intuitively.

src/routes/userRoutes.test.js
AliceHuang1027: agree that we needs to check both status code as well as the json-ed message

src/routes/userRoutes.test.js
AliceHuang1027: i think use mockFindOne is very precise ; however here there is a typo-- shall be mock not mcok ~~

AliceHuang1027: well, since you are being very consistent using mcok, that's good too ;)

src/routes/userRoutes.test.js
AliceHuang1027: yea, makes sense. btw do you expect to test logger info as well?

src/routes/userRoutes.test.js
kimong1: > We want `status` to be mocked as well. Because we want to see what argument is being passed to `status` function. Does it make sense?

This was a stupid comment. I didn't read your comment properly at that moment. Your argument is valid. Both way work well. I like mine because I think it's the way how actuall `app.status` is. We can do `app.json()` and `app.status().json()` which means `app.staus` returns `app` itself.

src/routes/userRoutes.test.js
kimong1: I fixed

src/routes/userRoutes.test.js
kimong1: I don't know actually cuz there is only vague idea what to include to the tests. Here is what I think how it is so far.

    1. Write tests for what your tested code is supposed to do
    2. Include tests the way your test covers 100% of tested file
    3. I don't think we should test every single helper function. If so, testing code can be unneccessarily too long and hard to be read

So I didn't expect logger

src/routes/userRoutes.js
```diff
@@ -35,5 +35,52 @@ routes.resetPassword = async (req, res) => {
...
+routes.createUser = async (req, res) => {
+  const userInfo = {
+      username: req.body.username,
+      email: req.body.email,
+      password: req.body.password
+  }
+  try {
+      await signUp(userInfo)
+      logger.info('Succeded creating user account', userInfo.username)
+      return res.status(200).json('Succeded creating user account')
```
greenc123, line 47: Try to send json with an object

```json({message: "Succeded creating user account"})```

File: src/routes/userRoutes.js [hunk 12]
```diff
@@ -35,5 +35,52 @@ routes.resetPassword = async (req, res) => {
...
+}
+
+routes.deleteUser = async (req, res) => {
+  if (!req.params.id) {
+      logger.info("user id was not provided")
+      return res.status(400).json({error: {message: "user id was not provided"}})
+  }
+  const {Accounts} = db.getModels()
+  try {
+      const account = await Accounts.findOne({
```
songz, line 61: this whole code block doesn't have good indentation. Could you indent please? Like `where` block below should be indented.

```
if (!account) {
  ...everything in here should be indented
}
```

src/routes/userRoutes.js
```diff
@@ -35,5 +35,52 @@ routes.resetPassword = async (req, res) => {
...
+      })
+      if (!account) {
+      logger.info("Cannot find user", req.params.id)
+      res.status(404).json({error: {message: "Cannot find user"}})
+      return
+      }
+      if (account.username !== req.session.username) {
+          logger.error("Username does not match to cookie", account.username, req.session.username)
+          res.status(500).json({error: {message: "Username does not match to cookie"}})
+          return
```
greenc123, line 74: for consistency could you 

```
return res.status(500).json(....)
``

File: src/routes/userRoutes.js [hunk 14]
```diff
@@ -35,5 +35,52 @@ routes.resetPassword = async (req, res) => {
...
+      }
+      await account.destroy()
+      logger.info('Succeded deleting user account', req.params.id)
+      return res.status(200).json('Succeded deleting user account')
+  } catch (err) {
+      logger.error("Deleting user failed", req.params.id, err)
+      res.status(500).json({error: {message: 'Deleting user failed. Please try again'}})
```
greenc123, line 81: you should send a 403 status code instead of a 500. There is nothing wrong with the server https://en.wikipedia.org/wiki/List_of_HTTP_status_codes

src/routes/userRoutes.js
```diff
@@ -35,5 +35,52 @@ routes.resetPassword = async (req, res) => {
...
+      }
+      })
+      if (!account) {
+      logger.info("Cannot find user", req.params.id)
+      res.status(404).json({error: {message: "Cannot find user"}})
+      return
+      }
+      if (account.username !== req.session.username) {
+          logger.error("Username does not match to cookie", account.username, req.session.username)
+          res.status(500).json({error: {message: "Username does not match to cookie"}})
```
greenc123, line 73: could you return a 403 this action should be forbidden by a user 
```
res.status(403)
```
500 means there is something wrong with the server https://en.wikipedia.org/wiki/List_of_HTTP_status_codes

src/routes/userRoutes.test.js
songz: this is like choosing between `1+2=3` or `2 + 1 = 3`. Just randomly pick one and stick to it :)

src/routes/userRoutes.test.js
songz: `logger` is best judgement for the developer to put. It doesn't need to be tested. It serves to be informational, not core functionality. Testing should cover core functionality, so logger should be tested only if its part of core functionality. LIke when a database module mess up, that would affect our production systems so maybe a logger there is necessary. There is no 100% rule for this.

PR #68, opened 2020-06-03
Send reset password [API], update README, update C.R.U.D. in tests

greenc123:


src/routes/router.js
```diff
@@ -0,0 +1,41 @@
+const express = require('express')
```
kimong1, line 1: We don't need this file

src/routes/userRoutes.js
```diff
@@ -0,0 +1,34 @@
...
+    where: {
+      email: email
+    }
+  })
+  if (!userAccount) {
+    return res.status(400).json({error: {message: "account does not exist"}})
+  }
+  logger.info(`user account found for user ${userAccount.id}, sending email now`)
+  try {
+    sendPasswordResetEmail(userAccount)
```
songz, line 24: await

src/routes/userRoutes.test.js
```diff
@@ -0,0 +1,46 @@
...
+  test('should throw error if req body does not have email', async () => {
+    const req = {
+      body: {}
+    }
+
+    await resetPassword(req, res)
+    expect(res.status.mock.calls[0][0]).toEqual(400)
+    expect(mockJson.mock.calls[0][0]).toEqual({error: {message: "invalid input"}})
+  })
+  test('should throw error if user account is not found', async () => {
```
songz, line 29: wrong description. Should send response with status 400 if user account does not exist

src/routes/userRoutes.test.js
```diff
@@ -0,0 +1,46 @@
...
+   const req = {
+     body: {
+       email: 'hello@world.com'
+     } 
+   } 
+
+    await resetPassword(req, res)
+    expect(res.status.mock.calls[0][0]).toEqual(400)
+    expect(mockJson.mock.calls[0][0]).toEqual({error: {message: "account does not exist"}})
+  })
```
songz, line 45: missing 2 more test cases:

1. should send success if send password is successful
2. should send 500 error if send password throws error

src/routes/userRoutes.js
```diff
@@ -0,0 +1,34 @@
...
+    }
+  })
+  if (!userAccount) {
+    return res.status(400).json({error: {message: "account does not exist"}})
+  }
+  logger.info(`user account found for user ${userAccount.id}, sending email now`)
+  try {
+    await sendPasswordResetEmail(userAccount) 
+    logger.info(`user resent password email sent to user ${userAccount.id}`)
+    return res.status(200).send('success')
```
songz, line 26: be consistent. If you are sending back `json` for all the other responses, you should send back a response here too. Instead, you are sending back a string.

src/routes/userRoutes.test.js
```diff
@@ -0,0 +1,93 @@
...
+
+    const req = {
+      body: {
+        email: 'hello@world.com' 
+      } 
+    }
+
+    await resetPassword(req, res)
+    expect(res.status.mock.calls[0][0]).toEqual(200)
+    expect(mockJson.mock.calls[0][0]).toEqual({error: {message: 'email sucessfully sent'}})
```
kimong1, line 91: Shouldn't it be `{sucess: {message: "Email sucessfully sent"}}`?

src/routes/userRoutes.js
```diff
@@ -0,0 +1,39 @@
...
+  if (!userAccount) {
+    return res.status(400).json({error: {message: "Account does not exist"}})
+  }
+
+  logger.info(`User account found for user ${userAccount.id}, sending email now`)
+  try {
+    await sendPasswordResetEmail(userAccount)
+
+    logger.info(`user reset password email sent to user ${userAccount.id}`)
+    return res.status(200).send({sucess: {message: "Email sucessfully sent"}})
```
kimong1, line 31: `send` -> `json`

src/routes/userRoutes.test.js
```diff
@@ -0,0 +1,93 @@
...
+
+   const req = {
+     body: {
+       email: 'hello@world.com',
+     } 
+   } 
+
+    await resetPassword(req, res)
+    expect(res.status.mock.calls[0][0]).toEqual(400)
+    expect(mockJson.mock.calls[0][0]).toEqual({error: {message: "account does not exist"}})
```
kimong1, line 44: image
This part failed cuz there's a typo in `account`

src/routes/userRoutes.test.js
```diff
@@ -0,0 +1,93 @@
...
+    })
+
+    const req = {
+      body: {
+        email: 'hello@world.com'
+      }
+    }
+
+    await resetPassword(req, res)
+    expect(res.status.mock.calls[0][0]).toEqual(500)
```
kimong1, line 65: image
It recieved `200` which means it didn't throw error
image
1. `resetPassword` doesn't throw any error so the description is wrong. It sends json object that has error property.
2. It sends error when `email` is falsey. Which means we should put no email value in `req.body` in order to recieve 400 error

kimong1: *In `userRoutes.js`
image
`resetPassword` sends `500` status when `sendPasswordResetEmail` throws error
*In `users.js`
image
`users.sendpasswordResetEmail` may throw error when `email.sendPasswordResetEmail` throws error.
(email is required from `mailer.js`)
*In mailer.js
image
`mgModule.sendPasswordResetEmail` will throw error when `mg.messages().send(data)` fails. `data` has `to` property whose value is `receiver` which is email address. So I assume you can make it throw error by passing a wrong email address

kimong1: Or you can simply mock `users.sendPasswordResetEmail` then do `sendPasswordResetEmail.mockImplement(() => { throw new Error('error') })`

src/routes/userRoutes.test.js
greenc123: yes

greenc123: good catch

PR #71, opened 2020-06-08
Login / Logout API

kimong1:
~~server.js did not pass 100% coverage~~
~~This uncovered code will be tested in another PR(#68 )~~
~~Conflicts will be resolved after the other PR is merged~~

Ready to merge! Please review

### DIFF
--- a/src/routes/userRoutes.js
+++ b/src/routes/userRoutes.js
@@ -1,30 +1,110 @@
+
+const {sendPasswordResetEmail, signUp, logIn} = require('../../lib/users')
 const logger = require('../../lib/log')(__filename)
 const db = require('../../sequelize/db')
-const {logIn} = require('../../lib/users')
+const routes = {}
 
-const obj = {}
+routes.resetPassword = async (req, res) => {
+  const email = req.body.email
+  logger.info('request received to send notification')
 
-obj.loginUser = async (req, res) => {
-    const inputInfo = {
+  if (!email) {
+    return res.status(400).json({error: {message: "invalid input"}})
+  }
+  
+  logger.info('user email is', email)
+  const {Accounts} = db.getModels()
+  const userAccount = await Accounts.findOne({
+    where: {
+      email: email
+    }
+  })
+
+  if (!userAccount) {
+    return res.status(400).json({error: {message: "Account does not exist"}})
+  }
+
+  logger.info(`User account found for user ${userAccount.id}, sending email now`)
+  try {
+    await sendPasswordResetEmail(userAccount)
+
+    logger.info(`user reset password email sent to user ${userAccount.id}`)
+    return res.status(200).json({success: {message: "Email sucessfully sent"}})
+  } catch (err) {
+    logger.error(`Could not send email to user ${userAccount.id}`)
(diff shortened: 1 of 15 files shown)


src/server.js
```diff
@@ -0,0 +1,48 @@
...
+        maxAge: 1000*60*5
+      }
+    }))
+
+    app.post('/api/session', loginUser)
+    app.delete('/api/session/:id', logoutUser)
+
+   server = app.listen(portNumber, () => {
+    resolve(app)
+    console.log(`Listening on portNumber ${portNumber}`)
```
greenc123, line 33: try to use logger.info

src/server.js
```diff
@@ -0,0 +1,48 @@
...
+    })
+  })
+}
+
+const stopServer = () => {
+  server.close()
+  console.log("The server has been closed")
```
greenc123, line 40: same thing here, logger instead of console log

src/userRoutes.test.js
```diff
@@ -0,0 +1,58 @@
...
+            throw new Error('Error')
+        })
+        const req = {
+            body: {
+                username: 'username',
+                email: 'em@i.l',
+                password: '1q2w3e4r'
+            }
+        }
+        await loginUser(req, res)
```
greenc123, line 26: test for the res.status also I think it should be a 500 if it fails

src/userRoutes.test.js
```diff
@@ -0,0 +1,58 @@
...
+        const req = {
+            body: {
+                username: 'username',
+                email: 'em@i.l',
+                password: '1q2w3e4r'
+            },
+            session: {}
+        }
+        await loginUser(req, res)
+        return expect(res.json.mock.calls[0][0]).toEqual('username is logged in')
```
greenc123, line 44: Same thing, I think it should be 200 if success

src/userRoutes.js
```diff
@@ -0,0 +1,30 @@
...
+    const inputInfo = {
+      username: req.body.username,
+      password: req.body.password,
+      email: req.body.email
+    }
+    try {
+      const userInfo = await logIn(inputInfo)
+      req.session.username = userInfo.username
+      logger.info('Logged in', userInfo.username)
+      res.status(200).json(`${userInfo.username} is logged in`)
```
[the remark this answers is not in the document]
kimong1: I think it doesn't matter for now. I don't know how this api will be used in FE so I just send a message

src/server.js
kimong1: I'm just gonna merge your server :) Thanks!

src/routes/userRoutes.js
```diff
@@ -77,9 +78,32 @@ routes.deleteUser = async (req, res) => {
       logger.info('Succeded deleting user account', req.params.id)
       return res.status(200).json('Succeded deleting user account')
```
greenc123, line 79: Could you send an object for your 200?

```
return res.status(200).json({success: {message: ''Succeded deleting user account}})

AliceHuang1027: In my opinion, this object is not necessary. Since success is success, do need to repeat what you succeed, it’s all in the description of the test. Well errors were generated for many reasons, that’s why we need to specify in the error message.

Issue #60, opened 2020-05-27
[API] - Update user


songz:
Updating db password

PATCH /api/users/:id

PR #72, opened 2020-06-08
 #60 Update User Database Password API

AliceHuang1027:
Done with 'updating db password' API 
PATCH /api/users/:id

Router file:
> data_router.js

Test file:
> userRoutes.test.js

Updated file:
>server.js 
>server.test.js


AliceHuang1027:
Screen Shot 2020-06-07 at 11 36 19 PM

PR #80, opened 2020-06-09
#60 Update DB Password Patch API

AliceHuang1027:
fixed the issues pointed by carlos and song.


services/__snapshots__/mailer.test.js.snap
```diff
@@ -0,0 +1,17 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
greenc123, line 1: This file seems like it should be on a frontend commit not for the API

src/userRoutes.test.js
```diff
@@ -0,0 +1,82 @@
...
+    beforeEach(() => {
+         jest.clearAllMocks()
+    })
+    
+    it('should send error if user id not found', async() => { 
+        const req ={
+            params:{id:null},
+            body:{password:null}
+         }
+         //req.params.id does not exist...
```
greenc123, line 32: You should take this comment out

src/userRoutes.test.js
```diff
@@ -0,0 +1,82 @@
...
+        const req = {
+            params:{id:12},
+            body:{password:'noexist'}
+        }
+        mockFindOne.mockReturnValue({
+            id:12
+        })
+        
+       users.setDBPassword.mockImplementation(() => {
+            console.log('mock it')
```
greenc123, line 72: You should take this console.log out

src/userRoutes.test.js
```diff
@@ -0,0 +1,82 @@
...
+            throw new Error('error')
+          })
+          await patch(req,res)
+          console.log(res.json.mock.calls[0][0])
```
greenc123, line 76: Same thing with this one

Issue #75, opened 2020-06-08
[FE] - Setup Integration test setup


songz:
1. create a route that goes to `/` and just renders welcome
2. create an integration test to start stop server and sends a request to `/` to match snapshot.
3. Setup view engine

PR #81, opened 2020-06-09
#75 [FE] Setup Integration test setup 

AliceHuang1027:
1. create a route in **server.js** that goes to / and just renders welcome in **welcome.ejs** 
2. create an integration test in **test/integration/welcome.test.js** to start stop server and sends a request to / to match snapshot.
3.Setup view engine: install ejs and app.set('view engine','ejs') 
4. changed part of server.test.js to test the app.set and app.get

Confusion: http://127.0.0.1 vs localhost  (since the former did not pass the test, the latter passed)
image is the former case (failed the test)
Screen Shot 2020-06-09 at 5 35 56 AM


src/server.js
```diff
@@ -13,6 +13,13 @@ const startServer = async (portNumber) => {
   await dbModule.start()
   return new Promise((resolve, reject) => {
     app = express()
+    app.set('view engine','ejs')
+    app.use(express.json())
+    app.get('/',(req,res)=>{
+      const user ={name:'donnell'}
```
greenc123, line 19: Who is donnell?

greenc123: Was this `user` object for testing purposes? It doesn't seem like it should be in a commit or code review file.

AliceHuang1027: Yea I was thinking whether to just take it out. However in that case we don’t need to use ejs

AliceHuang1027: > Who is donnell?

It’s one of my favorite comedians-Donnell Rawlings!

PR #82, opened 2020-06-10
[Backend function] reset user password

greenc123:
Creates a user reset password function that verifies to make sure the token is valid and then resets the password.

### DIFF
--- a/lib/users.js
+++ b/lib/users.js
@@ -110,7 +110,7 @@ users.resetUserPassword = async (token, password) => {
   if (!token) {
     logger.error('no token for resetUserPassword function')
     return new Error('no token for resetUserPassword')
-  } else if (!password) {
+  } if (!password) {
     logger.error('no password for resetUserPassword function')
     return new Error('no password for resetUserPassword')
   }
@@ -132,7 +132,9 @@ users.resetUserPassword = async (token, password) => {
     return new Error('user account not found')
   }
 
-  if (userAccount.tokenExpiration !== userToken) {
+  logger.info('found user account:', userAccount.id) 
+
+  if (userAccount.userToken !== userToken) {
     logger.error('user tokens do not match')
     return new Error('user tokens do not match')
   }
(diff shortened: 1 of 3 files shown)


lib/users.js
```diff
@@ -106,4 +106,51 @@ users.setDBPassword = async (userInfo, newPW) => {
   logger.info('DB password has set successfully', userInfo.username)
 }
 
+users.resetUserPassword = async (token, password) => {
+  if (!token) {
+    logger.error('no token for resetUserPassword function')
+    return new Error('no token for resetUserPassword')
+  } if (!password) {
```
songz, line 113: new line

lib/users.js
```diff
@@ -106,4 +106,51 @@ users.setDBPassword = async (userInfo, newPW) => {
...
+    logger.error('no password for resetUserPassword function')
+    return new Error('no password for resetUserPassword')
+  }
+
+  const unencryptedToken = Buffer.from(token, 'base64').toString('ascii')
```
songz, line 118: `toString()`

lib/users.js
```diff
@@ -106,4 +106,51 @@ users.setDBPassword = async (userInfo, newPW) => {
...
+  const account = JSON.parse(unencryptedToken)
```
songz, line 119: `const {userid, userToken} = JSON.parse( unencodedToken ) `

lib/users.js
songz: `unencodedToken`

lib/users.js
```diff
@@ -106,4 +106,51 @@ users.setDBPassword = async (userInfo, newPW) => {
...
+    return new Error('token is no longer valid')
+  }
+
+  const encryptedPassword = await bcrypt.hash(password, 10)
+
+  logger.info('Updating password for user:', userAccount.id)
+
+  await userAccount.update({
+    password: encryptedPassword
+  })
```
songz, line 153: `return userAccount`

PR #84, opened 2020-06-11
#75 [FE] Setup Integration test setup [Updated]

AliceHuang1027:
1. create a route in server.js that goes to / and just renders welcome in welcome.ejs
2. create an integration test in test/integration/welcome.test.js to start stop server and sends a request to / to match snapshot.
3. Setup view engine: install ejs and app.set('view engine','ejs') -- generate changes in package.json and package-lock.json (node-fetch and ejs installation)
changed part of server.test.js to test the app.set and app.get
Confusion: http://127.0.0.1 vs localhost (since the former did not pass the test, the latter passed)


tests/integration/welcome.test.js
```diff
@@ -0,0 +1,15 @@
+const {startServer,stopServer}=require('../../src/server')
+ const fetch = require('node-fetch')
+
+ describe('test welcome page',()=>{
+     beforeAll(async ()=>{
+         await startServer(20200)
```
songz, line 6: I actually cannot use `20200` on my computer. Could you make this configurable?

```
const testPort = process.env.TEST_PORT || 20200
const baseUrl = `http://localhost:${testPort}`
...
await startServer( testPort )
...
await fetch(baseUrl)
...
```

src/server.test.js
```diff
@@ -16,6 +16,7 @@ describe('Testing the server', () => {
   })
   test('getApp should return an express server after startServer has been called', async () => {
     express.mockReturnValue({
+      set:jest.fn(),
```
greenc123, line 19: Try to do what hoie did, he has a better practice to mock express once at the top of your file so your code is cleaner and easier to read.

```
const app = {
  set: jest.fn(),
  use: jest.fn(),
  post: jest.fn(),
  get: jest.fn()
}

express.mockReturnValue(app)

```

AliceHuang1027: Yea I was changing my code to meet the code in upstream server.js file, since I thought that was merged one. Will change back after Hoie’s code got merged.

tests/integration/welcome.test.js
AliceHuang1027: Ok

PR #86, opened 2020-06-12
Update User DB Password Patch API

AliceHuang1027:
#60  Update User DB Password Patch API [June, 11]
**server.js**
 add app.patch(...)
**server.test.js**
 add patch:jest.fn() to app and test of it
**userRoutes.js**
 add updateDBPassword function to routes object
**userRoutes.test.js**
 add updateDBPassword function tests


src/routes/userRoutes.js
```diff
@@ -81,6 +81,31 @@ routes.deleteUser = async (req, res) => {
       res.status(500).json({error: {message: 'Deleting user failed. Please try again'}})
   }
 }
+  routes.updateDBPassword = async(req,res) =>{
+    if(req.params.id===null ||!req.body.password ){
```
kimong1, line 85: I wonder if `req.params.id === null` is correct. If not passing any id parameter set `req.params.id` value as `null`, it makes sense. But what if it's `undefined` or empty string? We can do `!req.params.id` instead

src/routes/userRoutes.js
```diff
@@ -81,6 +81,31 @@ routes.deleteUser = async (req, res) => {
...
+    })
+    if(!userAccount){
+        logger.info(`account does not exist`)
+        res.status(400).json({error:{message:"account does not exist"}})
+        return
+    }
+    try {
+      await setDBPassword(userAccount,req.body.password) 
+      logger.info(`user ${userAccount.id} updates password`)
+      return res.status(200).json('success')
```
kimong1, line 103: `success` -> `{success: {message: 'success'}}`

src/routes/userRoutes.js
```diff
@@ -81,6 +81,31 @@ routes.deleteUser = async (req, res) => {
...
+    }
+    const {Accounts} = db.getModels()
+    const userAccount = await Accounts.findOne({
+        where:{
+          id:req.params.id
+        }
+    })
+    if(!userAccount){
+        logger.info(`account does not exist`)
+        res.status(400).json({error:{message:"account does not exist"}})
```
kimong1, line 97: `status(404)` for not found

src/routes/userRoutes.test.js
```diff
@@ -191,4 +191,75 @@ describe('Testing deleteUser function', () => {
...
-})
\ No newline at end of file
+})
+
+describe('testing upDBPassword function', () => {
+  beforeEach(() => {
+       jest.clearAllMocks()
+  })
+
+  it('should send 400 error if user id not found', async() => {
```
kimong1, line 201: Description should be `if input is invalid`

src/routes/userRoutes.test.js
```diff
@@ -191,4 +191,75 @@ describe('Testing deleteUser function', () => {
...
+      }
+
+      mockFindOne.mockReturnValue(userAccount)
+
+      const req = {
+          params:{id:12},
+          body:{password:'noexist'}
+      }
+
+
```
kimong1, line 254: Too many linebreaks!

src/routes/userRoutes.js
AliceHuang1027: So you think writing !req.params.id is better choice? In which case do you think it could be undefined?

src/routes/userRoutes.js
AliceHuang1027: I don’t think it’s necessary to complicate the message. Success is enough; at least I see most of the success messages in js5 class file is just ‘success’ in json . However error message usually tell you why it threw error.

src/routes/userRoutes.js
AliceHuang1027: 404 is page not found. Here is the account does not exist. 400 is bad request error; which we use here to indicate the request body id don’t match with any account in our database.

src/routes/userRoutes.test.js
AliceHuang1027: I assume invalid input means the input format is not correct; however here is more about we can’t find the account to match the req body id... so account does not exist cakes more sense to me

src/routes/userRoutes.test.js
AliceHuang1027: Okay; will try to squeeze them up. I was use Ling space to separate sections to make code easier to read and identify tho

src/routes/userRoutes.test.js
songz: then your expect message should match the description text right? its good to make them consistent.

src/server.js
```diff
@@ -26,7 +26,7 @@ const startServer = async (portNumber) => {
         maxAge: 1000*60*5
       }
     }))
-
+    app.patch('/api/users/:id',updateDBPassword)
```
songz, line 29: could you move this to the other `/api/users` routes so the bunch of them are close to each other?

src/routes/userRoutes.js
kimong1: I'd rather do !req.params.id because I think when frontend send request with no id parameter value, it will be undefined. I didn't tested so I don't know the answer. But !req.params.id works for both null and undefined so I'd rather do this way

src/routes/userRoutes.js
kimong1: You're right. Send actual data instead

src/routes/userRoutes.js
kimong1: I thought so too but I was convinced by these people


https://stackoverflow.com/questions/5604816/whats-the-most-appropriate-http-status-code-for-an-item-not-found-error-page

AliceHuang1027: In that case, shall we replace all the 400 we used in the code to 404 (say invalid input and account not found)? 

10.4.1 400 Bad Request

The request could not be understood by the server due to malformed syntax. The client SHOULD NOT repeat the request without modifications.


10.4.5 404 Not Found

The server has not found anything matching the Request-URI. No indication is given of whether the condition is temporary or permanent. The 410 (Gone) status code SHOULD be used if the server knows, through some internally configurable mechanism, that an old resource is permanently unavailable and has no forwarding address. This status code is commonly used when the server does not wish to reveal exactly why the request has been refused, or when no other response is applicable.

https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html

songz:
@AliceHuang1027 could you delete your `.DS_Store` file? It's something your computer created for your computer, we don't need that file in the codebase.

src/routes/userRoutes.js
songz: @AliceHuang1027 Could you return updated user object instead of a 'success' string? Here's an example: https://github.com/garageScript/databases/pull/89/files

AliceHuang1027:
> @AliceHuang1027 could you delete your `.DS_Store` file? It's something your computer created for your computer, we don't need that file in the codebase.

Damn it, let me check! I did not notice this...

src/routes/userRoutes.js
AliceHuang1027: Yea will fix it too

PR #89, opened 2020-06-13
Change API responses to returning resources

kimong1:
For example
image


lib/users.js
```diff
@@ -92,18 +92,20 @@ users.setDBPassword = async (userInfo, newPW) => {
...
       dbPassword: newPW
     }, {
       where: {
         username: userInfo.username
-      }
+      },
+      returning: true
     })
+    logger.info('DB password has set successfully', userInfo.username)
+    return updatedAccount[1][0]
```
greenc123, line 104: What does the updated account look like? Why are you only accessing index 1?

lib/users.test.js
```diff
@@ -96,32 +102,28 @@ describe('Sign up', () => {
...
-      password: 'password',
+      username: 'testuser',
+      password: 'abcd1234',
+      email: 'test.user@databases.com'
     }
-
-    await signUp(obj)
-
-    expect(mockCreateAccount.mock.calls.length).toEqual(1)
-    expect(mockCreateAccount.mock.calls[0][0].username).toEqual('Batman')
```
songz, line 121: so you decided to only check input and output of the function? You don't want to check if the `createAccount` is getting called?

lib/users.js
greenc123: I figured it out from the test but could you add a comment for future engineers?

lib/users.test.js
kimong1: This way makes sure creatAccount is called

lib/users.js
kimong1: Okay I'll do tonight

PR #90, opened 2020-06-14
[API] reset user password

greenc123:
Creates an API endpoint to call the userResetPassword function so users can reset their password


src/server.test.js
```diff
@@ -18,7 +20,9 @@ dbModule.close = jest.fn()
 const app = {
   set: ()=>{},
   use: () => {},
-  get: ()=>{},
+  get: (req, res)=>{
```
AliceHuang1027, line 23: btw is this necessary or get:()=>{} is just enough? 🤔

AliceHuang1027: otherwise we shall test this function later?

PR #91, opened 2020-06-15
codecov and github actions CI

greenc123:


kimong1:
Seems good to me. Once we figure out about integration test, then it'll be all good

PR #107, opened 2020-07-02
[FE] SetDBpassword

AliceHuang1027:
1. created setDBpassword.ejs
2. updated server.js by adding a get api
3. changed integration test file by adding the setDBpassword part


views/setDBpassword.ejs
```diff
@@ -0,0 +1,53 @@
...
+    .container {
+        width: 450px;
+        margin: auto;
+    }
+    .body{
+        margin:10px;
+        height:700px;
+        border: 1px solid rgba(0,0,0,0.5);
+        background:rgba(220,220,220,0.5)
+    }
```
greenc123, line 21: CSS should be handled with cdn

views/setDBpassword.ejs
```diff
@@ -0,0 +1,53 @@
...
+            },
+            body:JSON.stringify({
+                password:dbPassword.value
+            })
+        }).then(r=>r.json()).then(r=>{
+            return location.href = '/databases'
+        })
+    })
+</script>
+<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/open-fonts@1.1.1/fonts/inter.min.css">
```
greenc123, line 52: Move this to the top of the file so the page doesn't flash with unstyled content
https://www.techrepublic.com/blog/web-designer/how-to-prevent-flash-of-unstyled-content-on-your-websites/

PR #108, opened 2020-07-02
Front end sign up

greenc123:
* Signin links to signup page
* Adds integration tests

TODO:
* Needs to send email

### DIFF
--- a/views/signup.ejs
+++ b/views/signup.ejs
@@ -1,14 +1,34 @@
 <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/open-fonts@1.1.1/fonts/inter.min.css">
 <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@exampledev/new.css@1.1.2/new.min.css">
 
+
+<style>
+    * {
+        text-align: center;
+    }
+    input {
+        width: 400px;
+        height: 40px;
+        margin: 5px;
+    }
+    button {
+        width: 400px;
+        height: 40px;
+        margin: 10px;
+    }
+    .body {
+        margin: 10px;
+    }
+</style>
+
 <h1>Learn Databases</h1>
 
 <p>Create Account</p>
 
 <div>
   <input
-    class="square email"
-    type="text"
+    class="email"
+    type="email"
     placeholder="Email Address"
     required
     />
@@ -16,7 +36,7 @@
 
 <div>
   <input
-    class="square username"
+    class="username"
     type="text"
     placeholder="Username"
     required
@@ -24,7 +44,7 @@
 </div>
 
 <div>
-  <button class="square submit">Create Account</button>
+  <button class="submit">Create Account</button>
 </div>
 
 <div>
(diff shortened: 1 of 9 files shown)


views/signup.ejs
```diff
@@ -0,0 +1,53 @@
...
+    fetch('/api/users', {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json'
+      },
+      body: JSON.stringify({
+        username: username.value,
+        email: email.value
+      })
+    }).then(r => r.json()).then((r) => {
```
AliceHuang1027, line 49: （r）can just be written as () since r is not used...

views/signup.ejs
```diff
@@ -0,0 +1,53 @@
+<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/open-fonts@1.1.1/fonts/inter.min.css">
+<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@exampledev/new.css@1.1.2/new.min.css">
```
AliceHuang1027, line 2: i wonder if we need to adjust the margin for some of the div, input and button...

views/signup.ejs
greenc123: I'm going to send an email with the response

views/signup.ejs
```diff
@@ -0,0 +1,53 @@
...
+
+<h1>Learn Databases</h1>
+
+<p>Create Account</p>
+
+<div>
+  <input
+    class="square email"
+    type="text"
```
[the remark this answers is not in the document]
greenc123: include what?

views/signup.ejs
greenc123: maybe in the future but for right now the site doesn't have to look perfect

views/signup.ejs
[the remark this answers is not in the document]
greenc123: changed type to email

views/signup.ejs
```diff
@@ -0,0 +1,53 @@
+<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/open-fonts@1.1.1/fonts/inter.min.css">
+<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@exampledev/new.css@1.1.2/new.min.css">
+
+<h1>Learn Databases</h1>
+
+<p>Create Account</p>
+
+<div>
+  <input
+    class="square email"
```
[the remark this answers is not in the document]
greenc123: It was part of the old css

kimong1:
One thing needs to be added in sign up page. When users click `Create Account`, it redirects to sign in page immediately. I think it should show some message to tell users know what to do. For example:
```
submit.addEventListener('click', () => {
    ... ...
  alert("Check your email and set your password")
  return location.href = '/signin'
```

views/signup.ejs
songz: @greenc123 this actually redirects. This code here is the front end.

views/signup.ejs
songz: Will resolve in this issue: https://github.com/garageScript/databases/issues/110

Issue #112, opened 2020-07-06
[FE] - Reset Password Page


songz:
Afters receive a signup email, there is a link to take users to reset password page.

Users should be able to input their new password on this page to set their password.

PR #118, opened 2020-07-09
[FE] Set Password Page

kimong1:
#112 

* Check [here](https://kimong.learndatabases.dev/setPassword/token)

* Created page to path `/setPassword/:token`
    * Take password input twice, check if they match
    * Set user password
    * Automatically login and redirect to setDBPassword page
       
* Modified `userRoutes.resetPasswordEmail`
    * "Confirmation Email" -> "Set your password"

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -38,7 +38,7 @@ const startServer = async (portNumber) => {
     app.get('/signup', (req, res) => {
       res.render('signup')
     })
-    app.get('/setPassword/:id/:token', (req, res)=>{
+    app.get('/setPassword/:token', (req, res)=>{
       res.render('setPassword')
     })
     app.post('/api/notifications', resetPasswordEmail)
--- a/views/setPassword.ejs
+++ b/views/setPassword.ejs
@@ -38,20 +38,7 @@
             })
         }).then(r => r.json()).then((r) => {
             if (r.error) return alert(r.error.message)
-            fetch('/api/session', {
-                method: "POST",
-                headers: {
-                    'Content-Type': 'application/json'
-                },
-                body: JSON.stringify({
-                    username: r.username,
-                    password: r.password
-                })
-            }).then((r) => {
-                return location.href = '/setDBPassword'
-            }).catch((err) => {
-                alert("Password is set but login has failed. Please try again.", err)
-            })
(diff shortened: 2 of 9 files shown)


src/server.js
```diff
@@ -38,6 +38,9 @@ const startServer = async (portNumber) => {
     app.get('/signup', (req, res) => {
       res.render('signup')
     })
+    app.get('/setPassword/:id/:token', (req, res)=>{
```
songz, line 41: `userId` comes from the `token`. so `/setPassword/:token` is enough.

kimong1: How do you get id from token?

kimong1: Fixed

views/setPassword.ejs
```diff
@@ -0,0 +1,88 @@
...
+            headers: {
+                'Content-Type': 'application/json'
+            },
+            body: JSON.stringify({
+                token: token,
+                password: password1.value
+            })
+        }).then(r => r.json()).then((r) => {
+            if (r.error) return alert(r.error.message)
+            fetch('/api/session', {
```
songz, line 69: `return fetch(....)` You should not have nested promises. Promises goal is to not have these nested promises.

Isn't it better to do `location.href = '/'`?

kimong1: I think it's good to automatically log in the user when they set there password (I have lots of memory thinking that it's annoying that I have to tell the whole user information again just after I told them)
I can do this way
```
submit.addEventListener('click', asyc () => {
  try {
    fetch('/api/passwordReset', {...}).then(...)
  } catch {
  return
  }
  try {
    fetch('/api/session', {...}).then((r) => { location.href='/setDBPassword' })
  } catch {
  ...
  }
})
```

views/setPassword.ejs
```diff
@@ -0,0 +1,47 @@
...
+            <button class="submit">Submit</button>
+        </div>
+    </div>
+</main>
+
+<script>
+    const path = window.location.pathname.split('/')
+    const token = path.pop()
+    console.log('path is', path)
+    console.log('toekn is', token)
```
songz, line 20: remove `console.log` please.

kimong1: Done

PR #119, opened 2020-07-09
[FE] apply user session to updateDBPassword api

kimong1:
Actually, session middleware had been already set up. (Of course because we have login and logout api)
image

Because of the middleware above, on the server side, session data can be read/written simply like this:
image

So I just used this to `updateDBPassword` api.
Review please


src/routes/userRoutes.js
```diff
@@ -92,14 +92,14 @@ routes.loginUser = async (req, res) => {
   }
 }
   routes.updateDBPassword = async(req,res) =>{
-    if(!req.params.id ||!req.body.password ){
+    if(!req.session.username ||!req.body.password ){
```
greenc123, line 95: This is being a little nitpicky but try doing:

```
const username = req.session.username
const password = req.body.password
```

kimong1: Done!

src/routes/userRoutes.js
```diff
@@ -1,143 +1,178 @@
...
-      const updatedAccount = await setDBPassword(userAccount,req.body.password) 
-      logger.info(`user ${userAccount.id} updates password`)
-      return res.status(200).json({...updatedAccount.dataValues,password:null})
-    } catch (err) {
-      logger.error('Password update failed. Please try again',req.params.id, err)
-      return res.status(500).json({error: {message: 'Password update failed. Please try again'}})
-    }
+};
+routes.updateDBPassword = async (req, res) => {
+  const username = req.session.username;
```
kimong1, line 121: This is the part that is actually changed. Other changes are just formatting

kimong1:
This PR also include auto-formatting. I made a coment where is actually changed

PR #125, opened 2020-07-13
[FE] fixed the head.ejs

AliceHuang1027:
commit again and it shall fix the issue of include head.ejs in all the other files


views/partials/head.ejs
```diff
@@ -1,7 +1,6 @@
 <!--views/partials/head.ejs-->
 <meta charset='UTF-8'>
-<title>Welcome to learndb.dev </title>
-
+<title></title>
```
songz, line 3: couldy ou delete this tag if you don't need it?

AliceHuang1027: okay

AliceHuang1027:
Here are how the three pages look like (signin, signup, setDBpassword) :

Screen Shot 2020-07-13 at 4 29 38 PM
Screen Shot 2020-07-13 at 4 29 38 PM
Screen Shot 2020-07-13 at 4 30 45 PM

views/signup.ejs
```diff
@@ -1,26 +1,4 @@
...
-        width: 400px;
-        height: 40px;
-        margin: 10px;
-    }
-    .body {
-        margin: 10px;
-    }
-</style>
-
+<%- include('./partials/head') %>
```
AliceHuang1027, line 1: How to solve "EJS Unexpected token / - Using Partials" issue: https://cmsdk.com/node-js/ejs-unexpected-token---using-partials.html 👻

PR #131, opened 2020-07-16
Fix resetPW

kimong1:
Fixed setPassword page

It include auto-formatting
This part is only actually changed (just one single line)
https://github.com/garageScript/databases/pull/131/files#r455467649
User account data doesn't have `userid` key. It has `id` key

### DIFF
--- a/lib/users.js
+++ b/lib/users.js
@@ -1,159 +1,179 @@
-const logger = require('./log')(__filename)
-const yup = require('yup')
-const db = require('../sequelize/db')
-const bcrypt = require('bcrypt')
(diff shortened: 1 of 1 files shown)


lib/users.js
```diff
@@ -1,159 +1,179 @@
...
 
-  const {Accounts} = db.getModels()
+  const { Accounts } = db.getModels();
 
   const userAccount = await Accounts.findOne({
     where: {
-      userid: userid 
-    }
-  })
+      id: userid,
```
kimong1, line 147: This part is actually changed

Issue #128, opened 2020-07-15
[UI] - Ability to go to reset password page


songz:
As a user, when I forgot my password or username, I am unable to go to reset password page to reset my password. I don't know what to do or how to proceed.

Update
* signup page -> should have a link to reset password
* login page -> should say `reset password` instead of `find password`

PR #136, opened 2020-07-17
Add reset password page

kimong1:
Related issue : #128 [UI] - Ability to go to reset password page

This PR adds resetPassword page to the path `/resetPassword` click [here](https://kimong.learndatabases.dev/resetPassword) to see test website

It takes username or email.
When users click submit it checks the input and find user account data, send reset password email.
It alerts with the user's email address in case of when they know their username but not their email.
Users can check their email and go to the setPassword page with token.


src/routes/userRoutes.js
```diff
@@ -11,18 +11,18 @@ const routes = {};
...
   const email = req.body.email;
-  logger.info("request received to send notification");
-
-  if (!email) {
+  const username = req.body.username;
+  if (!email && !username) {
     return res.status(400).json({ error: { message: "invalid input" } });
   }
-  logger.info("user email is", email);
+  const filter = { where: {} };
```
songz, line 18: technically, this is called a `query`. You pass in a `query` into sequelize model's `findOne` function

kimong1: Fixed

PR #140, opened 2020-07-19
Fix login bug

kimong1:
There are some bugs where we can't log in. Don't worry I just fixed them all.

1. Log in is impossible because session middleware doesn't work
2. Log in is impossible **also** because we can't set password. 
    * Set password dosen't work because the column name where token is stored is wrong
3. After log in, it sends users to setDBPassword page even though DB password exists

In `users.test.js`, auto-formatting has changed all the lines.
I marked where are actually changed

### DIFF
--- a/lib/users.test.js
+++ b/lib/users.test.js
@@ -1,274 +1,278 @@
-jest.mock('mailgun-js')
-jest.mock('../sequelize/db')
-jest.mock('../services/mailer')
-jest.mock('./log')
+jest.mock("mailgun-js");
+jest.mock("../sequelize/db");
+jest.mock("../services/mailer");
+jest.mock("./log");
 // require('./log') and mock should be before requiring everything
 
-const logGen = require('./log')
+const logGen = require("./log");
 const logger = {
-    info: jest.fn(),
-    error: jest.fn()
-}
-logGen.mockReturnValue(logger)
+  info: jest.fn(),
+  error: jest.fn(),
+};
+logGen.mockReturnValue(logger);
 
-const {signUp, logIn, setDBPassword, sendPasswordResetEmail, resetUserPassword} = require('./users')
-const bcrypt = require('bcrypt')
-const crypto = require('crypto')
-const db = require('../sequelize/db')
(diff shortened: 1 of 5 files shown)


lib/users.test.js
```diff
@@ -1,274 +1,278 @@
...
+      userToken: "abc",
+    };
+    const token = Buffer.from(JSON.stringify(account)).toString("base64");
     mockFindOne.mockReturnValue({
       id: 2,
-      userToken: '234',
-    })
-
-    const password = '12345678910'
+      passwordReset: "234",
```
kimong1, line 237: actually changed

lib/users.test.js
```diff
@@ -1,274 +1,278 @@
...
-
-    const token = Buffer.from(JSON.stringify(account)).toString('base64')
-    const password = '12345678910'
+      userToken: "1234",
+    };
+    mockFindOne.mockReturnValue({
+      id: 3,
+      passwordReset: "1234",
+      tokenExpiration: Date.now() - 200,
+    });
```
kimong1, line 253: 247~253 is actually changed

lib/users.test.js
```diff
@@ -1,274 +1,278 @@
...
+    };
+    const token = Buffer.from(JSON.stringify(account)).toString("base64");
     mockFindOne.mockReturnValue({
       id: 2,
-      userToken: '234',
-    })
-
-    const password = '12345678910'
+      passwordReset: "234",
+    });
```
kimong1, line 238: 237 is actually changed

lib/users.test.js
```diff
@@ -1,274 +1,278 @@
...
+    };
+    const mockUpdate = jest.fn();
+    mockFindOne.mockReturnValue({
+      id: 4,
+      passwordReset: "1234",
+      tokenExpiration: Date.now() + 200,
+      update: mockUpdate,
+    });
+    const token = Buffer.from(JSON.stringify(account)).toString("base64");
+    const password = "12345678910";
```
kimong1, line 273: 263~273 is actually changed

Issue #134, opened 2020-07-16
[UI] - Ability to sign out


songz:
As a user, when I am using public computer, I should be able to sign out of my account when I am logged in.

PR #141, opened 2020-07-19
Add logout ability

kimong1:
Related issue: #134 [UI] - Ability to sign out

image

I put logout button to head
And added head to every page

Check [here](https://kimong.learndatabases.dev/)

* This PR include some code from [this PR](https://github.com/garageScript/databases/pull/140) which is not merged yet.

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -24,39 +24,40 @@ const startServer = async (portNumber) => {
   await dbModule.start();
   return new Promise((resolve, reject) => {
     app = express();
-    app.set("view engine", "ejs");
-    app.use(express.json());
     app.use(
       session({
         secret: process.env.SECRET,
         resave: false,
         saveUninitialized: true,
         cookie: {
-          secure: true,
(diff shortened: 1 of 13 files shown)


src/server.js
```diff
@@ -33,38 +33,40 @@ const startServer = async (portNumber) => {
...
+          secure: false,
           maxAge: 1000 * 60 * 5,
         },
       })
     );
+    app.set("view engine", "ejs");
+    app.use(express.json());
     app.get("/", (req, res) => {
-      res.render("welcome");
+      res.render("welcome", { username: req.session.username });
```
songz, line 44:
```
if (!req.session.username) {
  return res.render('welcome')
}
const user = await User.find({where: {username: req.session.username}})
if (!user) {
  return res.render('welcome')
}
if (!user.dbPassword) {
  return res.render('setDbPassword', {user})
}
return res.render('databases', {user})
```

Issue #144, opened 2020-07-20
[UI] - Render Postgres Tutorial MarkDown


songz:
Content: https://github.com/garageScript/databases/issues/97

Issue #97, opened 2020-06-22
[Content] Postgresql


songz:
# 0 Credentials
```
hostname: 165.227.55.105
username: account username
password: account setDbPassword value
```

or 

Create PG Credentials - Button

# 1 Intro
Postgres is an application. More specifically, a database application for you to organize your data. It can store data for you and you can have multiple apps connect to your Postgres:

![multiple apps connecting to postgres](https://mermaid.ink/img/eyJjb2RlIjoiZ3JhcGggTFJcblx0RFtBcHAxXSAtLT4gUG9zdGdyZXNcblx0RVtBcHAyXS0tPiBQb3N0Z3JlcyBcblx0RltBcHAzXSAtLT4gUG9zdGdyZXMiLCJtZXJtYWlkIjp7InRoZW1lIjoiZGVmYXVsdCJ9LCJ1cGRhdGVFZGl0b3IiOmZhbHNlfQ)

# 2 Basics
To quickly help you get started, we will go over a few basic commands that you can follow along. Explanations and best practices will be covered in the section below.

## 2.1 Connect
To connect to your app to Postgres, you need a module called `pg`. You need to specify where the postgres application is located (host and port), the name of your database, and your username / password to login to the database.

```
const { Client } = require('pg')

const startApp = async ()=>{
  client = new Client({
    host: '<host name>',
    port: 5432,
    user: '<username>',
    password: '<password>',
    database: '<database_name>'
  })
  await client.connect()
  console.log('Connected')
}

startApp()
```

## 2.2 Simple Commands
After connecting to Postgres, you can now send commands that Postgres can understand. These commands are called Sequel Query Language (SQL). Here are a few:

* `CREATE TABLE lesson( id serial PRIMARY KEY, title VARCHAR (256) );`
  * This creates a table called `lesson` that has 2 columns: `id` and `title`
  * `id` column is a primary key and the `serial` property means that it will automatically increase as you add rows into the table.
  * `VARCHAR (256)` means that title is a string of up to 256 characters.
  * **Make sure you don't run this over and over again. You only need to run this command once**
* `INSERT INTO lesson (title) VALUES ('postgres tutorial')`
  * This creates a new row in the `lesson` table with a title of 'postgres tutorial'.
* `select * from lesson where title = 'postgres tutorial'` 
  * This retrieves all the rows from the lesson table that has the title `"postgres tutorial'`.
* `UPDATE lesson SET title = 'postgres demo' WHERE id=1;`
[...]

PR #148, opened 2020-07-21
Add postgres tutorial page

kimong1:
* Related issue
  * #144 [UI] - Render Postgres Tutorial MarkDown
  * #97 [Content] Postgresql
* What this PR does
  * Postgres page is added to `/postgres` (Check [here](https://kimong.learndatabases.dev/postgres))
     
  * Databases page has buttons that send users to each database page. Path is changed from `/databases/postgres` to `/postgres` 
  * Previously we didn't separate web page rendering router functions since they are so simple. Now they can be complicated and long so I stored them in `/src/routes/renderRoutes.js`


kimong1:
It doesn't pass 100% coverage test. I'm still figuring it out

views/postgres.ejs
```diff
@@ -0,0 +1,194 @@
...
+<pre><code><span class="hljs-keyword">const</span> { Client } = <span class="hljs-built_in">require</span>(<span class="hljs-string">'pg'</span>)
+
+<span class="hljs-keyword">const</span> startApp = <span class="hljs-keyword">async</span> ()=&gt;{
+    client = <span class="hljs-keyword">new</span> Client({
+            <span class="hljs-attr">host</span>: <span class="hljs-string">'&lt;host name&gt;'</span>,
+                <span class="hljs-attr">port</span>: <span class="hljs-number">5432</span>,
+                    <span class="hljs-attr">user</span>: <span class="hljs-string">'&lt;username&gt;'</span>,
+                        <span class="hljs-attr">password</span>: <span class="hljs-string">'&lt;password&gt;'</span>,
+                            <span class="hljs-attr">database</span>: <span class="hljs-string">'&lt;database_name&gt;'</span>
+                              })
```
greenc123, line 24: Is there any way you can format this more so it's easier to read?

kimong1: I'm considering to use markedjs. I'll update soon

src/routes/renderRoutes.js
```diff
@@ -0,0 +1,18 @@
+const db = require("../../sequelize/db");
+const routes = {};
+
+routes.postgres = async (req, res) => {
+  if (!req.session.username) return res.redirect("/");
+  const { Accounts } = db.getModels();
+  const userAccount = await Accounts.findOne({
+    where: {
+      username: req.session.username,
```
greenc123, line 9: Can we look up by id here: `req.session.id`

kimong1: Oh yes

PR #152, opened 2020-07-21
Sign up errors

greenc123:
Screen Shot 2020-07-21 at 3 23 35 PM


kimong1:
> I see 1 line has changed. It's just removing logger. How does this fix the bug?

Issue #158, opened 2020-07-29
[UI] Remove databases view and add it to landing page


songz:
![Screen Shot 2020-07-29 at 10 03 43 AM](https://user-images.githubusercontent.com/686933/88830480-d9e77600-d182-11ea-9223-ac588df76ad0.png)

remove `/databases` route, it is not necessary.

PR #159, opened 2020-07-29
#158 [UI] Remove databases view and add it to landing page

greenc123:
### What this PR does:
- removes the previous content from /databases and adds it to the landing page
- redirects signedin users to postgres landing page


greenc123:
Screen Shot 2020-07-29 at 4 39 44 PM
Screen Shot 2020-07-29 at 4 39 58 PM
Screen Shot 2020-07-29 at 4 38 08 PM

views/welcome.ejs
```diff
@@ -20,6 +37,35 @@
       <p>Neo4j, a Graph Database. <a href="#">Neo4j</a></p>
       </blockquote>
 
+
```
songz: could you remove the blockquotes above?

src/server.js
```diff
@@ -57,13 +58,12 @@ const startServer = async (portNumber) => {
...
       res.render("setPassword", { username: req.session.username });
     });
-    app.get("/databases", (req, res) => {
-      res.render("databases", { username: req.session.username });
-    });
     app.get("/resetPassword", (req, res) => {
       res.render("resetPassword", { username: req.session.username });
     });
     app.get("/postgres", postgres);
+    app.get("/mongodb", mongodb);
```
songz, line 65: rather than doing this, it should be refactored to 

```
const {database} = require('./routes/renderRoutes')
...
app.get("/database", database)
```

src/routes/renderRoutes.js
```diff
@@ -15,4 +15,12 @@ routes.postgres = async (req, res) => {
   });
 };
 
+routes.mongodb = async (req, res) => {
```
songz, line 18: comment below, we only want 1. `routes.database`

views/welcome.ejs
```diff
@@ -34,5 +67,22 @@
 
     <p>We collect anonymous analytics using Simple Analytics. View live: <a href="#">simpleanalytics.com/learndatabases.dev</a></p>
 
+      <script>
+  const postgres = document.querySelector('.postgres')
+  const mongodb = document.querySelector('.mongodb')
+  const neo4j = document.querySelector('.neo4j')
+
+  postgres.addEventListener('click', () => {
```
kimong1, line 75: I just tested this page. These buttons are not valid because....
1. If users are not logged in, nothing happens because they will be redirected back to `"/"`
2. If users are logged in, they can't see these buttons because theses buttons won't be rendered.
    * Right now, `postgres` page is rendered when the user is logged in

src/server.js
```diff
@@ -69,6 +64,8 @@ const startServer = async (portNumber) => {
     app.delete("/api/session", logoutUser);
     app.post("/api/passwordReset", userResetPassword);
 
+    app.get("/:database", postgres);
```
kimong1, line 67: `postgres` router can't render postgres page because `postgres.ejs` is missing. I think you changed the name to `database.ejs`

views/welcome.ejs
```diff
@@ -1,24 +1,57 @@
...
+    margin-bottom: 15px;
+    padding: 10px;
+    padding-bottom: 20px;
+    border-radius: 5pX;
+    cursor: pointer;
+  }
+
+  .db-card:hover {
+    border: solid 1px rgb(10, 70, 120);
+  }
```
kimong1, line 18: These css makes it look like the cards are clickable but no script is listening to clicks.
I think we need to decide first if we are gonna keep these buttons or not

src/routes/renderRoutes.js
```diff
@@ -30,7 +30,9 @@ routes.landingpage = async (req, res) => {
   if (!user.dbPassword) {
     return res.render("setDBpassword", { username: req.session.username });
   }
-  res.render("databases", { username: req.session.username });
+  res.render("welcome", {
+    username: req.session.username,
+  });
```
kimong1, line 35: Changing this means that now users can't go to databases page... I think we need alternative feature that allows users to see databases page. How about redirecting users to databases page when they click database buttons?

songz:
Closing issue.

PR #170, opened 2020-08-05
Integrate databases page to landing page

kimong1:
1. Related Issue
   * [UI] Remove databases view and add it to landing page #158
2. What this PR does
   * Integrates `databases.ejs` to `welcome.ejs`
   * Changes redirecting path `"/databases"` to `"/"` (all of them)
   * Landing page no longer has to decide what to render. So this PR removes `landingpage` function from `renderRoutes.js`
   * [Preview](https://kimong.learndatabases.dev/)
     image
3. ~~What I need help with~~
   * ~~It can't run integration test~~
   * ~~I can connect to the page and test login, cliking buttons - so it should be able to run integration test but why no??~~


views/welcome.ejs
```diff
@@ -8,16 +8,17 @@
...
-
-      <blockquote>
-      <p>Postgres, an SQL Database. <a href="#">PostGresSQL</a></p>
-      </blockquote> 
-
-      <blockquote>
-      <p>Neo4j, a Graph Database. <a href="#">Neo4j</a></p>
+      <blockquote class="db-card mongodb">
+        <h3 class="db-title">MongoDB</h3>
+        <p class="db-discription">An open source database management system that uses a document-oriented database model which supports various forms of data.<p>
```
anthonykhoa, line 17: I think

`An open source database management system that uses a document-oriented database model which supports various forms of data.`

would be better written as

`An open source database management system using a document-oriented database model that supports various forms of data.`

views/welcome.ejs
```diff
@@ -8,16 +8,17 @@
...
+      </blockquote>
+      <blockquote class="db-card neo4j">
+        <h3 class="db-title">Neo4j</h3>
+        <p class="db-discription">A graph database management system that is the most papular graph database and the 22nd most popular database overall.</p>
```
anthonykhoa, line 21: `papular` should be `popular`

views/welcome.ejs
kimong1: Nice

views/welcome.ejs
kimong1: Nice

views/welcome.ejs
```diff
@@ -35,4 +36,29 @@
     <p>We collect anonymous analytics using Simple Analytics. View live: <a href="#">simpleanalytics.com/learndatabases.dev</a></p>
 
   </body>
+
+  <style>
```
danielthalhuber, line 40: I was about to ask why this `style` element is not in the `head` or `body`. But I scrolled up a little further and now I see that we're including `./partials/head`, which already contains a complete `html` element.

Since we have another `html` element in _this_ template, the resulting HTML to contains two separate `html` elements. That partials template contains many elements that normally would be placed in the head, outside of the `html` element. The resulting HTML source is confusing but somehow parsed by modern browsers.

There are a lot of good opportunities to fix and streamline the code here. But this otherwise solid PR shouldn't be held up by a preexisting condition. This `style` element can be addressed later.

Issue #171 contains a record of the view template issues mentioned here.

kimong1: Nice comment. I didn't realize it. I'll fix soon.

songz: nice find @danielthalhuber !

songz: thank you for filing the issue.

Issue #166, opened 2020-08-03
[UX] - Take out username from signup flow


songz:
as a user, I should not have to put in my username when I signup. I always get confused and end up tying my password instead.

PR #172, opened 2020-08-06
Take out username from signup&login process

hoiekim:
# 1. Related Issue
* [UX] - Take out username from signup flow #166
# 2. What this PR does
* Backend
   * Replaces input field `username` to `email` (or removes them) from `userRoutes.js`
   * Replaces input field `username` to `email` (or removes them) from `users.js`
   * Changes sequelize setting: Username can be null now
* Frontend ([Preview](https://kimong.learndatabases.dev/))
   * Takes out `username` input field from fronted
      * For Example 
image
   * Changes head message from "Welcome `username`!" to "Welcome `email`!"
      * For Example
image

### DIFF
--- a/sequelize/db.js
+++ b/sequelize/db.js
@@ -1,69 +1,74 @@
-const { Sequelize, DataTypes } = require('sequelize')
-const logger = require('../lib/log')(__filename)
-require('dotenv').config()
+const { Sequelize, DataTypes } = require("sequelize");
+const logger = require("../lib/log")(__filename);
+require("dotenv").config();
 
-let sequelize
-const dbModule = {}
+let sequelize;
+const dbModule = {};
 
-const models = {}
+const models = {};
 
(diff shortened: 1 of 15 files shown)


sequelize/db.js
```diff
@@ -1,69 +1,74 @@
...
       }
-    })
-    models.Accounts = await sequelize.define('account', {
+    );
+    models.Accounts = await sequelize.define("account", {
       username: {
         type: DataTypes.STRING,
         unique: true,
-        allowNull: false
+        allowNull: true,
```
anthonykhoa, line 31: Is it possible to just altogether remove `username` as an attribute?
If we are no longer using `username` and it will not have any significance, then I think we should remove the `username` attribute.

hoiekim: We are gonna discuss about it soon

PR #174, opened 2020-08-12
closes #173 - Connect to neo4j database

### LINKED ISSUE #173: Connect to neo4j database
A neo4j database has been set up on another server for us to use. The codebase should be updated such that the server connects to the neo4j database on server startup.

anthonykhoa:
A neo4j database has been set up on another server for us to use. The codebase should be updated such that the server connects to the neo4j database on server startup. Another issue this PR solves is not having a `uniform` coding style for the codebase.

This PR will:
* Connect to the neo4j database set up on another server
* Update package.json and package-lock.json files
* Create a `.prettierrc.js` file to maintain a consistent coding style across codebase


database/neo4j/neo4j.test.js
```diff
@@ -0,0 +1,10 @@
+let { closeNeo4j } = require('./neo4j')
+
+describe('Neo4j database', () => {
+	closeNeo4j = jest.fn().mockReturnValue(Promise.resolve())
+	let driver.close = jest.fn().mockReturnValue(1)
+  it('it should call end when closing neo4j', async () => {
+    await closeNeo4j()
+    expect(closeNeo4j).toHaveBeenCalledTimes(1)
```
hoiekim, line 8: I think expecting `closeNeo4j` to have been called is weird.

In `neo4js.js`, when you run `closeNeo4j` function, it runs `driver.close` function. and `driver` object is the return of `neo4j.driver()`. So why don't you mock `neo4j-driver` instead of `closeNeo4j`?

src/server.js
```diff
@@ -1,8 +1,8 @@
-const express = require("express");
```
hoiekim, line 1: I think we should stop using auto-formatting....
Otherwise we will keep changing the whole code every time we take turns to edit the same files

anthonykhoa: we should have a linter that can autoformat for us.

database/neo4j/neo4j.test.js
```diff
@@ -0,0 +1,19 @@
...
+  //needed for closeNeo4j
+  const driver = neo4j.driver()
+
+  beforeEach(startNeo4j)
+
+  driver.close = jest.fn().mockReturnValue(Promise.resolve())
+
+  test('Should resolve when closeNeo4j is called', async () => {
+    const res = await closeNeo4j()
+    expect(res).resolves
```
hoiekim, line 17: I may be wrong, but I think this test code has some problem like this:

The test might pass because what you expect is just to resolve. But what you did in 6th~13th line won't affect anything to this 17th line which actually tests something. I think you tried to get the functions that `closeNeo4j` calls. But what you actually did is creating new functions/objects with the same name. You should mock them before you require `'./neo4j'`.
How about changing `expect(res).resolves` to `expect(driver.close).toHaveBeenCalled` to make sure if the test code works?

anthonykhoa: I tried doing that before and couldn't get any tests to pass by using 
`expect(driver.close).toHaveBeenCalledTimes(1)`
Screen Shot 2020-08-13 at 12 48 38 AM

anthonykhoa: hmm i have some ideas, i will try them out

anthonykhoa: ah yea `jest.mock('neo4j-driver')` is what i wwas missing

hoiekim (approved):

PR #175, opened 2020-08-13
Bug: user should be logged in after setting pw

hoiekim:
I missed `req.session.email = account.email` in `userResetPassword` function.
This PR fixes it.


src/routes/userRoutes.js
```diff
@@ -162,6 +162,7 @@ routes.userResetPassword = async (req, res) => {
     const account = await resetUserPassword(token, password);
     logger.info("User password reset for", account.email);
     req.session.userid = account.id;
+    req.session.email = account.email;
```
songz, line 165:
```
routes.userResetPassword( req, res)
expect( req.session).toEqual({...})
```

Issue #167, opened 2020-08-03
[UX] - Create Postgres Button to create credentials if users does not have a database.


songz:
It is important for users to know what is happening, so the click is important.

Users should click on the create postgres button to create a postgres account.

PR #176, opened 2020-08-14
Add create db button

hoiekim:
# 1. Realted Issue
* [UX] - Create Postgres Button to create credentials if users does not have a database. #167

# 2. What this PR does
* It adds `Create my database` button on `postgres` page
* It edits `pg.js` file to do these
   * Since we use email as username, it has `"@"` in its value. But username for postgres db should be alphabetical
   * So it should take out `"@"` and `"."` from the email string and use them as db username

# 3. Preview
[click here](https://kimong.learndatabases.dev/postgres)
image

### DIFF
--- a/database/postgres/pg.js
+++ b/database/postgres/pg.js
@@ -22,21 +22,22 @@ pgModule.closePGDB = () => {
 
 pgModule.createPgAccount = async (username, password) => {
   if (!username || !password) return;
+  const usernameAlphabetical = username.replace("@", "").replace(".", "");
   try {
     // Could not escape user input by using $1 $2
     //   https://github.com/brianc/node-postgres/issues/539
 
-    const sqlQuery1 = escape(`CREATE DATABASE %s;`, username);
+    const sqlQuery1 = escape(`CREATE DATABASE %s;`, usernameAlphabetical);
     const sqlQuery2 = escape(
       `create user %s with encrypted password %Q`,
-      username,
+      usernameAlphabetical,
       password
     );
 
     const sqlQuery3 = escape(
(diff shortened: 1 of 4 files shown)


public/lessons/postgres.md
```diff
@@ -12,75 +21,75 @@ To connect to your app to Postgres, you need a module called `pg`. You need to s
...
+  await client.connect()
+  console.log('Connected')
 }
 
 startApp()
-  ```
+```
 
 ## 2.2 Simple Commands
   After connecting to Postgres, you can now send commands that Postgres can understand. These commands are called Sequel Query Language (SQL). Here are a few:
```
aortizoj15, line 40: I believe it should be Structured Query Language instead of Sequel Query Language

database/postgres/pg.js
```diff
@@ -22,41 +22,45 @@ pgModule.closePGDB = () => {
 
 pgModule.createPgAccount = async (username, password) => {
   if (!username || !password) return;
+  const usernameAlphabetical = username.replace("@", "").replace(".", "");
```
anthonykhoa, line 25: Why is the `replace(".", "")` needed?

anthonykhoa: maybe we should auto generate the username using a library that can compose an alphanumeric string for us.

songz:
```
song@mail.zheng.club
song.mail@zheng.club
```

Build your own encoder? https://github.com/samswen/base62-util/blob/master/src/index.js#L4

songz:
tech debt: https://github.com/garageScript/databases/issues/178

Issue #164, opened 2020-08-02
[UX] - Click on create Postgres Account should show the user credentials


greenc123:
## What this PR does:

Create a function that checks the user's PG account to see if it exists.

songz:
done

PR #182, opened 2020-08-23
#164 userHasPgAccount

greenc123:
# What this PR does:
- Creates a function to check is a user has a Postgres account


database/postgres/pg.js
```diff
@@ -61,4 +61,22 @@ pgModule.deletePgAccount = async (username) => {
   }
 };
 
+pgModule.userHasPgAccount = async (username) => {
+  if (!username) return false;
+  const usernameAlphabetical = username.replace("@", "").replace(".", "");
```
hoiekim, line 66: https://github.com/garageScript/databases/pull/179/files

In this PR, I changed how we encode username. So we don't use `usernameAlphabetical` variable anymore. You can simply put `username`

greenc123:
No longer can ssh into the c0d3 server. New PR here: https://github.com/garageScript/databases/pull/183

PR #183, opened 2020-08-24
#164 userHasPgAccount function

greenc123:
# What this PR does

- returns a boolean to check if the user has a Postgres account.
- Update Jest
- Some tests were not completed so now all databases are being closed


database/postgres/pg.js
```diff
@@ -61,4 +60,18 @@ pgModule.deletePgAccount = async (username) => {
...
 };
 
+pgModule.userHasPgAccount = async username => {
+  if (!username) return false;
+
+  try {
+    const user = await escape(`SELECT 1 FROM `, username, ` WHERE `, username);
+
+    const result = await client.query(user);
+    return result ? true : false;
```
hoiekim, line 70: It can be just `!!result`

database/postgres/pg.js
```diff
@@ -61,4 +60,18 @@ pgModule.deletePgAccount = async (username) => {
   }
 };
 
+pgModule.userHasPgAccount = async username => {
+  if (!username) return false;
+
+  try {
+    const user = await escape(`SELECT 1 FROM `, username, ` WHERE `, username);
```
songz, line 67: this query doesn't look right.

songz: please test your code! There's no way this could work. https://stackoverflow.com/questions/8546759/how-to-check-if-a-postgres-user-exists

songz (approved):

hoiekim (approved):

anthonykhoa (approved):

PR #192, opened 2020-09-01
Add setTimeout in integration test

hoiekim:
## Issue
image
* Sometimes it takes too long when the integration test starts the test server so it gets this error

## What this PR does
* It adds `jest.setTimeout(10000)` to `welcome.test.js` file. Then it worked okay with me :)

### DIFF
--- a/tests/integration/welcome.test.js
+++ b/tests/integration/welcome.test.js
@@ -1,6 +1,9 @@
+jest.setTimeout(10000);
+
 const session = require("express-session");
 const { startServer, stopServer } = require("../../src/server");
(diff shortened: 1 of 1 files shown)


tests/integration/welcome.test.js
```diff
@@ -1,6 +1,9 @@
+jest.setTimeout(10000);
+
 const session = require("express-session");
 const { startServer, stopServer } = require("../../src/server");
 const fetch = require("node-fetch");
+require("dotenv").config();
```
hoiekim, line 6: Also somebody forgot to put this

Issue #188, opened 2020-08-31
UX - Check for user account before display credentials


songz:
<img width="1378" alt="Screen Shot 2020-08-31 at 3 44 53 PM" src="https://user-images.githubusercontent.com/686933/91775883-1bfc4100-eba1-11ea-91c1-55db893bb344.png">

PR #197, opened 2020-09-06
Check for user account before display credentials in Postgres page

hoiekim:
## Related issue
* UX - Check for user account before display credentials #188

## What this PR does
* `renderRoutes.postgres` function will check if user has created database, and send `dbExists` variable as boolean
* Depends on `dbExists` value, `postgres.ejs` page decides to show create button or credentials information.
* Mostly copied from [this PR](https://github.com/garageScript/databases/pull/194)
* `pg.userHasPgAccount` function is still in progress(look this [PR](https://github.com/garageScript/databases/pull/183)) so I wrote a temporary function for it.

## Preview
image

### DIFF
--- a/views/postgres.ejs
+++ b/views/postgres.ejs
@@ -1,19 +1,63 @@
 <%- include("./partials/head") %>
+<h3>Learn PostgreSQL</h3>
+<div id="introduction">
+</div>
 <div id="content"></div>
 <script src="https://cdn.jsdelivr.net/npm/marked/marked.min.js"></script>
 <script>
+  const username = "<%= username %>"
+  const dbPassword = "<%= dbPassword %>"
+  const dbExists = <%= dbExists %>
+
+  const credentials = document.createElement("pre")
+  credentials.innerHTML = `<code>host: 165.227.55.105 
+username: ${username}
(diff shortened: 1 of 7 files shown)


views/postgres.ejs
```diff
@@ -1,19 +1,74 @@
...
-  const createDb = () => {
-    fetch('/api/createDatabase', { method: "POST" }).then(r => r.json()).then((r) => {
-      if (r.error) return alert(r.error.message)
-      return alert(r.success.message)
-    }) 
-  }
 </script>
+
+<style>
+  #button {
```
songz, line 66: for future, please update '#button' to something better. there's a couple things wrong:

1. button is too generic.
2. applying an `id` to a generic element is not good.

views/postgres.ejs
```diff
@@ -1,19 +1,74 @@
 <%- include("./partials/head") %>
+<h3>Learn PostgreSQL</h3>
+<div id="introduction">
```
songz, line 3: could we have renamed this to `database.ejs` to be reusable for other db as well?

greenc123 (approved):

songz (approved):

PR #225, opened 2020-09-25
Integrates databases rendering process

hoiekim:
## Issue
* Inspired by [this comment](https://github.com/garageScript/databases/pull/197#discussion_r484099867)
* `postgres.ejs` and `elastic.ejs` has similar structure. We can make them into a single file with some logic that decides which database page to render and use this file for other databases too
## What it does
* Integrates `postgres.ejs` and `elastic.ejs` into `tutorial.ejs`
   * In `tutorial.ejs`, it takes in database name parameter and decide which `.md` file to request
* Integrates `postgres` and `elastic` functions into `database` function in `renderRoutes.js` file
   * It checks `req.params.database` value and put them as a parameter when it renders `tutorial.ejs`
   * Also decides which database to check if the user has created their db according to `req.params.database`
* Changes path from `"/postgres"`, `"/elasticsearch"` to `"/tutorial/:database"`
#### before
image
#### after
image


src/routes/renderRoutes.js
```diff
@@ -4,52 +4,39 @@ const pg = require("../../database/postgres/pg");
...
+      where: {
+        id: req.session.userid,
+      },
     });
+    if (req.params.database === "Postgres") {
+      renderData.dbExists = await pg.userHasPgAccount(userAccount.username);
+    }
+    if (req.params.database === "Elasticsearch") {
+      renderData.dbExists = await es.checkAccount(userAccount);
+    }
```
anthonykhoa, line 33: Lines 28-33 and lines 15-20 can be combined together

src/routes/userRoutes.js
```diff
@@ -164,13 +164,13 @@ routes.createDatabase = async (req, res) => {
   }
 
   try {
-    if (req.params.database === "postgres") {
+    if (req.params.database === "Postgres") {
       await pgModule.createPgAccount(username, dbPassword);
       return res
         .status(200)
```
anthonykhoa, line 170: You didn't write this, but can you fix this? The `.status(200)` is unneeded because the status defaults to 200.

anthonykhoa: Also on line 175 too :)

views/tutorial.ejs
```diff
@@ -11,28 +11,29 @@
   integrity="sha512-r3qr7vMdHYmWUzqC7l4H/W03ikAyiFIQld+B71yAmHhu7wcsnvm/adhikM+FzDxxK9ewznhCVnAm+yYZpPBnSg=="
   crossorigin="anonymous" />
 <script>
```
anthonykhoa, line 13: There seems to be a missing `hljs.initHighlightingOnLoad()` here.

src/routes/renderRoutes.js
hoiekim: No they can't. Because 15-20 is for both logged in / not logged in situation while 28-33 is only for logged in situation

views/tutorial.ejs
hoiekim: I took it out because I thought it's not necessary. It only makes this part looks weird.
image
Other parts are still highlighted because of line 71.

anthonykhoa (approved):

songz (approved):

PR #228, opened 2020-09-27
Implement anonymous user's database creation

hoiekim:
#### Feedback from HN
image
#### Application plan
* Anonymous user: Provide temporary database with expiration after a certain period of time (Anonymous user means user account that has no email data)
* Authenticated user: Provide database that is permanently usable

## 2. What it changes
* Modify `users.signUp` function in `users.js` file
    * So that it can create user account without email data.
* Modify `userRoutes.createDatabase` function in `userRoutes.js` file
    * If user is not logged in, It creates user account using `users.signUp` function
    * It sends response with JSON object of the new created user account data
* Modify `createDb` function and button on `tutorial.ejs` page
   * It fetches request to `"api/createDatabase/:database"` and render created credentials

## 3. ETC.
* Automatic expiration of temporary database is not setup in this PR. Since it's becoming a pretty large change / addition, I think it's better to cover expiration part in another PR.


lib/users.js
```diff
@@ -38,35 +38,22 @@ users.sendPasswordResetEmail = async (userAccount) => {
 };
 
 users.signUp = async (userInfo) => {
-  try {
```
hoiekim, line 41: I took out `try ... catch` here because the whole `signUp` function goes in another `try ... catch` in `userRoutes` function

lib/users.js
```diff
@@ -38,35 +38,22 @@ users.sendPasswordResetEmail = async (userAccount) => {
...
-    await schema.validate(userInfo);
-  } catch (err) {
-    throw new Error(err);
-  }
-
-  const { email } = userInfo;
-  const { Accounts } = db.getModels();
-  const userAccount = await Accounts.findOne({
```
hoiekim, line 49: I also took this part out because `Accounts.create` function in 49th(before 64th) line also check unique-validation.

lib/users.js
```diff
@@ -38,35 +38,22 @@ users.sendPasswordResetEmail = async (userAccount) => {
...
-    where: {
-      email: email,
-    },
-  });
+  await schema.validate(userInfo);
   const username = uniqueNamesGenerator({
     length: 2,
-    dictionaries: [adjectives, colors, animals],
```
hoiekim, line 56: I took `colors` out because it only uses the first 2 dictionaries since the length is 2

songz (approved):

views/postgres.ejs
```diff
@@ -65,6 +68,39 @@ database: ${username}</code>`
...
+    fetch("/api/anonymous/Postgres", { method: "POST" }).then((r) => r.json()).then((r) => {
+      console.log(r)
+      if (r.error) {
+        const message = res.error.message;
+        alert(message);
+        disabled = false
+        button.classList.remove("disabled")
+        button.innerText = "Create my database"
+      } else {
+        credentials.innerHTML = `<code>host: learndatabases.dev
```
anthonykhoa, line 86: Why is this different from createdb function? Is it possible to use the introduction.append function?
Also, instead of 2 different createdb functions, I think it would make sense to combine them and create just 1 createdb function. And then the back-end can check whether to make a `anon` or `non-anon` 

Or another idea I can think of that is better, is to have the fetch decide which URL to hit. Depending on whether the user is logged in or not.

hoiekim: That's absolutely true. Even the API can be integrated. So I changed it again. Thanks for feedback

songz (approved):

anthonykhoa (approved):

PR #229, opened 2020-09-27
closes #221 - Update Readme.md

### LINKED ISSUE #221: README.md is outdated
It's not possible to start server locally without valid credentials right now. Attempt to run it results in error `password authentication failed for user \"uliss\""`

Ulisseus:
**What does this PR do?**
The main idea is to make life for new students easier.

Readme has proper step by step installation guide now. Github workflow guide probably should be kept separately and updated [there](https://github.com/garageScript/curriculum/wiki/Engineering-Workflow) if needed since it's useful for new students across all repos. 

I added comments to env.example to briefly explain what each variable does. I removed unused DIALECT variable and added a new one NODE_ENV (String: "production"/"development") which switches links in confirmation emails from learndatabases.dev to your localhost. 

/lib/getEnv.js was removed since it duplicates `require("dotenv").config()` functionality. Added some tests to mailer.test.js and updated its snapshot.


lib/users.js
```diff
@@ -29,7 +29,7 @@ users.sendPasswordResetEmail = async (userAccount) => {
   const accountJSON = JSON.stringify(account);
   const encodedToken = Buffer.from(accountJSON).toString("base64");
 
-  await email.sendPasswordResetEmail(userAccount.email, encodedToken);
+  await email.sendPasswordResetEmail(userAccount.email, encodedToken, process.env.NODE_ENV, process.env.PORT);
```
songz, line 32: This is going to break production. Production environments usually use a proxy and this app runs on port xxxx, which will be passed in even though publicly the port should be 443. 

A better approach is for `sendPasswordResetEmail` to use a `HOSTNAME` process.env variable. For example, one of our engs has `ldb.hoie.kim` as their development environment.

services/mailer.js
```diff
@@ -1,13 +1,18 @@
 const mailgun = require("mailgun-js");
 const logger = require("../lib/log")(__filename);
-const getEnvVar = require("../lib/getEnvVar");
```
hoiekim, line 3: I can't remember why I did this..
I thought the reason was because we had to hijack the env variable when we test or something. So I reviewed the code again but it doesn't seem that way. So weird!

Ulisseus: I thought it had something to do with tests too after I tried to add new ones to mailer. With separate getEnv you could check env variables separately and without it I need to pass them as arguments to email.sendPasswordResetEmail.

Ulisseus:
I squashed my local commits to make history look nice, but local and origin paths diverged and now my log history looks even worse with a lot of unnecessary merges/fixes. I wonder how to do it properly.

hoiekim (approved):

services/mailer.js
```diff
@@ -1,13 +1,16 @@
...
+  apiKey: process.env.MAILGUN_API_KEY,
+  domain: process.env.MAILGUN_DOMAIN,
+});
 
 const mgModule = {};
 
-mgModule.sendPasswordResetEmail = (receiver, token) => {
-  const link = `https://learndatabases.dev/setPassword/${token}`;
+mgModule.sendPasswordResetEmail = (receiver, token, hostname = "https://learndatabases.dev") => {
+  const link = `${hostname}/setPassword/${token}`
```
songz, line 13: instead of argument, why not use `process.env` directly? This way you don't have to change the function shape. On line 6 and 7 the `env` function is used already.

songz (approved):

PR #231, opened 2020-09-29
Automatically cleaning expired anonymous user accounts

hoiekim:
#### See also
Implement anonymous user's database creation #228
#### Feedback from HN
image
#### Application plan
* Anonymous user: Provide temporary database with expiration after a certain period of time (Anonymous user means user account that has no email data)
* Authenticated user: Provide database that is permanently usable

## 2. What it changes
* Add `lib/util.js` file
    * It has `cleanAnonymous` function.
    * This function queries expired user accounts and delete their postgres and elasticsearch database as well as the user account itself
    * It runs again in every 24 hours
* Modify `server.js` file
    * When server starts, it runs `cleanAnonymous` function


lib/util.test.js
```diff
@@ -0,0 +1,58 @@
...
+  it("should call findAll with correct queries", async () => {
+    Accounts.findAll.mockReturnValue([]);
+    await util.cleanAnonymous();
+    const args = Accounts.findAll.mock.calls[0][0];
+    expect(args.attributes[0]).toEqual("id");
+    expect(args.attributes[1]).toEqual("username");
+    expect(Number(args.where.and[0].createdAt.lt)).toBeLessThan(
+      Number(new Date(new Date() - 5 * 24 * 60 * 60 * 1000 + 1))
+    );
+    expect(args.where.and[1].email).toBeNull();
```
hoiekim, line 43: I wrote some strict test here because I thought that it's important to protect this query. Someone can accidently clear up the whole db by making a little change.

lib/util.js
```diff
@@ -0,0 +1,39 @@
...
+            createdAt: {
+              [Op.lt]: new Date(new Date() - 5 * 24 * 60 * 60 * 1000),
+            },
+          },
+          { email: null },
+        ],
+      },
+    });
+    logger.info(`${userAccount.length} expired accounts are found`);
+    userAccount.forEach(async (e) => {
```
anthonykhoa, line 27: Using async await in a `forEach` loop won't work. This will cause multiple asynchronous operations to run, but the problem is that those operations won't be guaranteed to finish before line 34(`logger.info`) gets read.

This stackOverflow answer has really good explanation and solutions for this case.
https://stackoverflow.com/questions/37576685/using-async-await-with-a-foreach-loop?rq=1

hoiekim: That's a really good feedback. I'm fixing it soon

lib/util.test.js
songz: NICE. this should be in the comment. 

```
// Hi future engineers, I wrote some strict test here because I thought that it's important to protect this query. Someone can accidently clear up the whole db by making a little change. If you are reading this, please pass down this knowledge
```

greenc123 (approved):

lib/util.js
```diff
@@ -0,0 +1,49 @@
...
+            },
+          },
+          { email: null },
+        ],
+      },
+    });
+    logger.info(`${userAccount.length} expired accounts are found`);
+    return Promise.all(
+      userAccount.map(async (e) => {
+        const pgDbExists = await pg.userHasPgAccount(e.username);
```
anthonykhoa, line 29: Will lines 29-32 be guaranteed to be called in sequential order?

hoiekim: Since the callback is `async`, the `await`s in front of each functions will guarantee that those functions run sequentially in my opinion.

anthonykhoa (approved):

songz (approved):

PR #233, opened 2020-10-02
[DNM] closes #232 - Arango functions to stop/start server, and create/delete account

### LINKED ISSUE #232: arangodb functions
startDB, closeDB, createAccount and deleteAccount functions should be created for arrango database.

anthonykhoa:
Functions to start and stop a connection to an Arango database, as well as delete and create users from an Arango database, need to be created. These functions are needed before creating the API routes that will call those functions.

This PR will:
* Create a `sendFetch` helper function to use in the back end.
* In the `database/elastic/elastic.js` file, `sendESRequest`is instead replaced with the `sendFetch` function. 
* Create functions to start and stop a connection to a Arango Database
* Create functions to delete and create users on a Arango Database
* Delete all neo4j files


greenc123 (approved):

lib/sendFetch.js
```diff
@@ -0,0 +1,17 @@
...
+const sendFetch = (path, method, body, authorization) => {
+  if (!path) return;
+  const options = {
+    method,
+    headers: {
+      "content-type": "application/json",
+    },
+  };
+  body && (options.body = JSON.stringify(body));
+  authorization && (options.headers.authorization = authorization);
```
hoiekim, line 12: Does this mean.. `if (authorization) { options.headers.authorization = authorization }` ?

database/arango/arango.js
```diff
@@ -0,0 +1,78 @@
...
+const logger = require("./../../lib/log")(__filename);
+
+const arangoModule = {};
+let db;
+
+arangoModule.checkIfDatabaseExists = async (name) => {
+  // returns array of objects containing a '_name' key.
+  // Ex [{ _name: 'lol' }, { _name: 'cakes' }]
+  const names = await db.databases();
+  return names.map((db) => db._name).includes(name);
```
hoiekim, line 13: It seems that this function requests all of the data from the database in JSON form and then search for the name. I think this can cause some inefficiency. Just like postgres and elasticsearch, I assume that there is a way to query the databases with specific name using arango.

database/arango/arango.js
```diff
@@ -0,0 +1,78 @@
...
+arangoModule.createAccount = async (account) => {
+  if (!account) return;
+  const { username, dbPassword } = account;
+  if (!username || !dbPassword) return;
+  if (await arangoModule.checkIfDatabaseExists(username)) return;
+
+  try {
+    await db.createDatabase(username, {
+      users: [{ username, password: dbPassword }],
+    });
```
hoiekim, line 38: How about putting `log.info("Created ArangoDB account")`? Not only this one but also other functions

lib/sendFetch.js
anthonykhoa: Yeah

database/arango/arango.js
anthonykhoa: Ah yes there is, I considered doing it that way, but eventually concluded that it is better to stick to using their arangojs API.

Another way I could do this is by making a request to a http REST api that arango provides. This is how deleting an account is done -- I use their http REST API. Here are the pros and cons of using each method to check if a user exists:

# Using http REST api:
* Pros
    1) Queries for only one user, so when the promise is received in `const names = await db.databases()`, I don't have to use `map` and `includes` functions.
* Cons
    1) Have to make 2 requests to an arango database -- the first time to get a jwt token, the second time to check if a user exists. This is a type of additional logic that arangojs api would not need.

# Using arangojs api:
* Pros
    1) Only 1 request is made to an arango database -- getting a jwt is not necessary because we authenticated ourselves in the function to connect to the arango database.
    2) Keeps the code consistent - the functions to stop and start a connection to an arango database are done using their arangojs api, so using their arangojs api to check if a user exists is appropriate. 
* Cons
    1) Queries for more than one user, so additional logic must be performed to check if a user exists, after receiving the promise (`using .map and .includes`). Using http REST api, I would not need this type of additional logic.


# Efficiency
I believe there is no noticeable difference in efficiency between the two methods, unless learndatabases.dev gets a looot of users. You can [see here](https://www.arangodb.com/2018/02/nosql-performance-benchmark-2018-mongodb-postgresql-orientdb-neo4j-arangodb/) that on average, it took arango only `1.07 seconds` to aggregate a collection of `1,632,803` documents. Since efficiency is not a concern, I think using arangojs is better for now. In the future, if efficiency becomes a concern, then we can switch to using the http REST api that arango provides.

database/arango/arango.js
anthonykhoa: Will do. Thanks!

database/arango/arango.js
hoiekim: As you said it won't be a big problem for now since we don't have a lot of requests but I want our app to be open to a chance to be so.

I have done some research and found something that can be a clue.

1. https://arangodb.github.io/arangojs/latest/classes/_database_.database.html#query
This function let us use AQL
2. https://www.arangodb.com/docs/stable/aql/tutorial-filter.html
This is how to get document that matches specific condition

So in this case I assume it should be like this:
```
const dbExists await !!db.query(aql`
  FOR d IN database
  FILTER d._name == ${name}
  RETURN d.name
`);
return dbExists
```

anthonykhoa: I tried looking into using `AQL` commands to do user management, but I wasn't able to do so. This is why I used http rest API to delete arango users, because I couldn't use `AQL` commands to do so. I think it is possible for us to make a collection of users to query from, but that would make our code more complex than it has to be, and so we should not do that.

hoiekim (approved):

database/arango/arango.js
songz: https://github.com/garageScript/databases/issues/237

PR #244, opened 2020-10-28
closes #53 - add influxDB

### LINKED ISSUE #53: [New DB] - Time Series DB
https://docs.timescale.com/latest/getting-started/installation/ubuntu/installation-apt-ubuntu

Ulisseus:
InfluxDB was chosen over TimescaleDB mostly because latter requires older versions of Postgres and maintaining multiply psqls on the same machine is rather annoying. It's also more popular at least for now. 

Since this PR adds new influxDB button on welcome page snapshots were updated. 

To safely merge:
- [InfluxDB 1.8](https://docs.influxdata.com/influxdb/v1.8/introduction/install/) needs to be installed
- INF_HOST and INF_PORT environment variables should be configured, by default it's `localhost` and `8086`
- src/routes/renderRoutes.js line 22   ` Influx: "PRODUCTION INFLUXDB HOST",` needs to be changed to whatever host will be used in production

Proofreading of influxDB tutorial is appreciated.


database/influx/influx.js
```diff
@@ -0,0 +1,61 @@
...
+    await client.grantPrivilege(username, "WRITE", username);
+    logger.info(
+      `Sucessfully created new user and database with ${username} name`
+    );
+  } catch (err) {
+    logger.error(err);
+    throw new Error(`failed to create new influx user with ${username} name`);
+  }
+};
+influxModule.deleteAccount = async (account) => {
```
anthonykhoa, line 32: This function only uses `username`, so why is the entire `account` object needed?

lib/util.js
```diff
@@ -37,6 +38,9 @@ util.cleanAnonymous = async () => {
         const arangoDbExists = await arango.checkIfDatabaseExists(username);
         if (arangoDbExists) await arango.deleteAccount(username);
 
+        const influxDbExists = await influx.checkAccount(username);
+        if (influxDbExists) await influx.deleteAccount(user);
```
anthonykhoa, line 42: You have the choice to pass in `username`, so you should do that instead of `user`.  I don't think it makes sense to pass in an object to a function that only uses one `string data-type` variable as an input. That is confusing -- when I look at the function parameters, it makes me think the function needs something else from `user` other than just `user.username`

src/server.js
```diff
@@ -19,6 +19,7 @@ let server = null;
 let app = null;
 
 const arangoModule = require("../database/arango/arango");
+const influxModule = require("../database/influx/influx.js");
```
anthonykhoa, line 22: You don't need the `.js` when importing a JS file

lib/util.js
Ulisseus: Basically, this code threw me off. PG and Elastic use username for checks and user for deletions while Arango needs only username in both cases. I was trying to be consistent so I added user as argument for Influx. 
```
    const pgDbExists = await pg.userHasPgAccount(username);
    if (pgDbExists) await pg.deletePgAccount(user);

    const esDbExists = await es.checkAccount(username);
    if (esDbExists) await es.deleteAccount(user);

    const arangoDbExists = await arango.checkIfDatabaseExists(username);
    if (arangoDbExists) await arango.deleteAccount(username);
```
I changed it, so deleteAccount now needs only username as argument.

PR #247, opened 2020-11-16
Closes #161 - add Ignite key-value in-memory database

### LINKED ISSUE #161: Implement Redis Module [Key-Value]
Documentation
https://redis.io/

npm
https://www.npmjs.com/package/express-redis-cache

`startRedisDB` function must start the database
`closeMongoDB` function must close the database
`createMongoDB` function must create an account.
`deleteMongoDB` function must delete an account.

Ulisseus:
While Redis supports creating username-password pairs with Access Control List most node.js clients still don't have this feature. Some clients allow you to authorize from node, but creating and deleting users still needs to be done via redis-cli. So Redis is not an option. The same applies to Memcached and probably most other in-memory databases. Since these databases were designed as caches their security is either non-existent or they support only one global password. 

To safely merge:
* Apache Ignite needs to be [installed](https://ignite.apache.org/docs/latest/quick-start/nodejs) (requires JDK)
* `IGNITE_HOST`, `IGNITE_USER` and `IGNITE_PASSWORD` should be added to .env file. Default user-password is `ignite`, host is `127.0.0.1:10800`

Start the server with provided configuration `bash ignite.sh path-to-igniteConfig.xml`. It's configured to have persistence (if server runs out of memory heap file on disk will be used, it's also a requirement for authentication for some reason) and to limit in-memory region to 200mb. 

Tutorial is pretty barebones so far, additions are welcome.

### DIFF
--- a/database/ignite/ignite.js
+++ b/database/ignite/ignite.js
@@ -1,0 +1,76 @@
+const IgniteClient = require("apache-ignite-client");
+const logger = require("./../../lib/log")(__filename);
+const { SqlFieldsQuery } = require("apache-ignite-client/lib/Query");
+const IgniteClientConfiguration = IgniteClient.IgniteClientConfiguration;
+const CacheConfiguration = IgniteClient.CacheConfiguration;
+require("dotenv").config();
+let igniteModule = {};
+let igniteClient;
+igniteModule.startIgnite = async () => {
+  try {
+    igniteClient = new IgniteClient();
+    const configuration = new IgniteClientConfiguration(process.env.IGNITE_HOST)
+      .setUserName(process.env.IGNITE_USER)
+      .setPassword(process.env.IGNITE_PASSWORD);
+    await igniteClient.connect(configuration);
+    logger.info("Connected to Ignite database");
+  } catch (err) {
+    logger.error(err);
+    throw new Error(err);
+  }
+};
+igniteModule.checkAccount = async (username) => {
+  if (!username) return;
+  try {
+    const caches = await igniteClient.cacheNames();
+    return caches.includes(username);
+  } catch (err) {
+    logger.error(`Error checking ignite database for ${username}`);
+    throw new Error(err);
+  }
+};
+igniteModule.closeIgnite = async () => {
+  try {
+    await igniteClient.disconnect();
+    logger.info("Closed connection to Ignite database");
+  } catch (err) {
+    logger.error(`Error closing connection to Ignite database`);
+    throw new Error(err);
+  }
+};
+
+igniteModule.createAccount = async (user) => {
+  const { username, dbPassword } = user;
+  if (!username || !dbPassword) return;
+  try {
+    const cache = await igniteClient.getOrCreateCache(
+      username,
+      new CacheConfiguration().setSqlSchema("PUBLIC")
+    );
+    await cache.query(
+      new SqlFieldsQuery(
+        `CREATE USER "${username}" WITH PASSWORD '${dbPassword}'`
+      )
+    );
+    logger.info(`Successfully created ${username} cache`);
+  } catch (err) {
+    logger.error(`Error creating ignite ${username}`);
+    throw new Error(err);
+  }
+};
+igniteModule.deleteAccount = async (username) => {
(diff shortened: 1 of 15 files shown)


src/routes/userRoutes.js
```diff
@@ -143,6 +144,7 @@ const createDatabaseAccount = {
   Postgres: pgModule.userHasPgAccount,
   Elasticsearch: es.createAccount,
   Arango: arangoModule.createAccount,
+  Ignite:igniteModule.createAccount
```
anthonykhoa, line 147: There is a missing space between `:` and `igniteModule.createAccount`

src/routes/renderRoutes.js
```diff
@@ -10,19 +11,22 @@ const dev_dbHost = {
...
   Arango: process.env.ARANGO_URL,
+  Ignite:process.env.IGNITE_HOST
 };
 
 // This is the 'host' url for a person's database credentials in prod.
 const dbHost = {
   Postgres: "learndatabases.dev",
   Elasticsearch: "elastic.learndatabases.dev",
   Arango: "arangodb.learndatabases.dev",
+  Ignite:"ADD_IGNITE_HOST"
```
anthonykhoa, line 22: There is a missing space between `:` and `ADD_IGNITE_HOST`

src/routes/renderRoutes.js
```diff
@@ -10,19 +11,22 @@ const dev_dbHost = {
   Postgres: process.env.HOST,
   Elasticsearch: process.env.ES_HOST,
   Arango: process.env.ARANGO_URL,
+  Ignite:process.env.IGNITE_HOST
```
anthonykhoa, line 14: There is a missing space between `:` and `process.env.IGNITE_HOST`

lib/util.test.js
```diff
@@ -58,16 +61,19 @@ describe("Testing cleanAnonymous function", () => {
...
     expect(es.deleteAccount).not.toHaveBeenCalled();
     expect(arango.deleteAccount).not.toHaveBeenCalled();
+    expect(ignite.deleteAccount).not.toHaveBeenCalled();
   });
   it("should call database functions if expired accounts are found", async () => {
     Accounts.findAll.mockReturnValue([{ destroy: () => {} }]);
     pg.userHasPgAccount = () => true;
     es.checkAccount = () => true;
     arango.checkIfDatabaseExists = () => true;
+    ignite.checkAccount = ()=>true;
```
anthonykhoa, line 71: Can you change the format to include spaces?

`ignite.checkAccount = () => true`

lib/util.js
```diff
@@ -37,6 +38,9 @@ util.cleanAnonymous = async () => {
         const arangoDbExists = await arango.checkIfDatabaseExists(username);
         if (arangoDbExists) await arango.deleteAccount(username);
 
+        const igniteDbExists = await ignite.checkAccount(username);
+        if(igniteDbExists) await ignite.deleteAccount(username);
```
anthonykhoa, line 42: Can you add a space after the `if`? For codebase styling consistency. We have a `prettierrc.js` file that is supposed to take care of styling. Are you using prettier?

https://github.com/garageScript/databases/blob/master/.prettierrc.js

database/ignite/ignite.js
```diff
@@ -0,0 +1,76 @@
+const IgniteClient = require("apache-ignite-client");
+const logger = require("./../../lib/log")(__filename);
+const { SqlFieldsQuery } = require("apache-ignite-client/lib/Query");
+const IgniteClientConfiguration = IgniteClient.IgniteClientConfiguration;
+const CacheConfiguration = IgniteClient.CacheConfiguration;
+require("dotenv").config();
+let igniteModule = {};
```
anthonykhoa, line 7: `igniteModule` should be declared `const` instead of `let`

lib/util.js
Ulisseus: I use prettier as default vscode formatter usually, so I forgot to check these files before submitting. I run prettier over all database files in new commit and it picked up a few inconsistencies here and there, mostly empty spaces around objects. 
Everything should be fixed now.

